PR #5, opened 2023-08-18
html-css/Halima

halima770:


halima770:
homework

index.html
```diff
@@ -0,0 +1,77 @@
...
+    <nav>
+        <div class="logo">
+      <p class="log" >portfo<span>lio</span></p>  
+        </div>
+        <ul>
+        <li><a href="#">Home</a></li> 
+        <li> <a href="#">About</a></li>
+        <li><a href="#">Skills</a></li>
+        <li><a href="#">Projects</a></li>
+        <li><a href="#">Contact</a></li>
```
mehmetnuricavdar, line 21: Well written 👏🏼 but links should lead to different parts of the web page.

index.html
```diff
@@ -0,0 +1,77 @@
...
+                
+                <li>
+                    <i class="fa fa-linkedin" aria-hidden="true"></i>
+                    <a href="https://github.com/halima770"> github</a>
+                </li>
+                
+                </ul>
+              </section>
+     
+              <section id="portfolio">
```
mehmetnuricavdar, line 68: Why section has 'id' here while the other sections have 'class' ?  I recommend avoiding using 'id' as the best practice is using 'class'

style.css
```diff
@@ -0,0 +1,315 @@
...
+    margin: 0px;
+    padding: 0px;
+    box-sizing: border-box;
+    
+}  
+
+
+
+
+
```
mehmetnuricavdar, line 23: I suggest formatting the code with the VS Code extension Prettier for clean code since there are many empty lines.

style.css
```diff
@@ -0,0 +1,315 @@
...
+nav{
+    width: 100%;
+   /* height: 75px; */
+    /*line-height: 75px;*/
```
mehmetnuricavdar, line 27: I guess you were tying the styles but i recommend before submitting the code delete this kind of comment to keep your code clean.

index.html
```diff
@@ -0,0 +1,77 @@
...
+                <h2>Portfolio</h2>
+                <div class="project">
+                  <h3>FCC Survey Form</h3>
+             <!---  <p>Description of the project...</p> -->
+                  <a href="https://www.freecodecamp.org/learn/2022/responsive-web-design/build-a-survey-form-project/build-a-survey-form">View Project</a>
+                </div>
```
mehmetnuricavdar, line 74: Footer?

style.css
```diff
@@ -0,0 +1,315 @@
...
+
+span{
+    color: gold;
+}
+
+header{
+    /*text-align: center;*/
+   /* padding: 80px ; */
+    /*padding 4% 0; */
+
```
mehmetnuricavdar, line 91: Same here for deleting unnecessary code before submit.

index.html
```diff
@@ -0,0 +1,77 @@
...
+                <li>java Script</li>
+                <li>java</li>
+            </ul>
+            </section> 
+
+            <section class="contact-info">
+                <h2>Contact information</h2>
+                <ul>
+                    <li>
+                        <i class="fa fa-linkedin" aria-hidden="true"></i>
```
mehmetnuricavdar, line 52: Can you tell me why you used 'aria-hidden'

style.css
```diff
@@ -0,0 +1,315 @@
...
+    box-shadow:0 0 10px rgba(250, 247, 247, 0.963); 
+    cursor: pointer;
+    
+}
+
+nav .logo p{
+    font-size: 30px;
+    font-weight: bold;
+    float: left;
+    color:white;
```
mehmetnuricavdar, line 48: using color codes is better for the future and get used to them from the beginning

style.css
```diff
@@ -0,0 +1,315 @@
...
+    }
+}
+
+@media screen and (max-width: 768px) {
+    #portfolio {
+        top: auto;
+        left: auto;
+        transform: none;
+    }
+}
```
mehmetnuricavdar, line 295: You could use only one '@media screen and (max-width: 768px)' and then put styles under it since it is for the same size.

mehmetnuricavdar (changes requested):
Welcome to HYF! It seems you understand concepts! 
There are some improvements that are needed and it is completely normal, no worries about that. 

First of all, you need to check this HTML/CSS homework folder since it is not in the correct folder as it is inside the general repository.
While I was checking it wasn't deployed to Netlify. You need to do that as well.
You should check styles since about/contact/skills are intertwined.
You added responsiveness but it doesn't work as expected since when most of the content is getting lost you can check it with Chrome inspect.
if you need help you can use support channels on Slack.

halima770:
Hello Mehment, thank you for the suggestions you provided. I've implemented the changes you mentioned. However, I'm still experiencing issues with the website's responsiveness. Could you please assist me with addressing this problem?

mehmetnuricavdar:
Hi Halima, to be able to do your website for different size of screens you and see how ti looks you can use 'inspect' of the browser and set the screen size as desired then you can adjust your styles. Also you can see how i did in the screen video i posted. You may wanna check https://web.dev/learn/design/media-queries/ or ask #html-css-support on Slack.

PR #16, opened 2023-08-19
html-css/kateryna

katookap:


mehmetnuricavdar (approved):
Hi Kateryna, it is a really nice job, very beautiful design and well written. You used HTML/CSS concepts very well. 

The only thing worth mentioning is that I couldn't understand why you used 'id' since you only used it for the 'footer paragraph' which you could do with '.footer p'
It is not recommended to use 'id' in HTML unless you really need it for unique elements that require specific styling or for JS interactions. 

Other than very well done! Keep rocking it! Looking forward to seeing more of your projects! 🚀⭐️

katookap:
> Hi Kateryna, it is a really nice job, very beautiful design and well written. You used HTML/CSS concepts very well.
> 
> The only thing worth mentioning is that I couldn't understand why you used 'id' since you only used it for the 'footer paragraph' which you could do with '.footer p' It is not recommended to use 'id' in HTML unless you really need it for unique elements that require specific styling or for JS interactions. The bonus: you can add your Netlify link to the Readme file.
> 
> Other than very well done! Keep rocking it! Looking forward to seeing more of your projects! 🚀⭐️

Thank you very much for the review! Did put my heart and soul into the project. As for the "id", I think I didn't find any other way in my brain at the time how to do it differently. Will correct asap:)

PR #19, opened 2023-08-20
html-css/rumana-aafrin

rumana-aafrin:
I have also uploaded images that I have used in my html


portfolio.css
```diff
@@ -0,0 +1,574 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  margin-right: 10px;
+}
+
+.initial {
+  font-size: 48px;
+  color: var(--portfolio-text, #090808);
+}
```
shpomp, line 57: It would be good to have more descriptive classes. If I just look at the CSS file and see "circle" or "initial", I have no idea what in the page it could be and have to go and search for those classes in HTML. It is the best practice to have class names that provide more information at first glance, at the same time still keeping them simple. 
`circle` could be `nav-logo` or `nav-round-logo`
`initial` could be `nav-logo-letter`
This applies to more of your classes, I just took a few as examples 🙏🏽

portfolio.css
```diff
@@ -0,0 +1,574 @@
...
+
+  .footer {
+    padding: 10px;
+  }
+
+  .footer-text,
+  .linkedin-text {
+    font-size: 18px;
+  }
+}
```
shpomp, line 573: There is a lot of CSS for a page that is relatively simple. 
That is more CSS than sometimes a multi-page app could have. 
This is all natural - it always happens in the beginning when first learning CSS. 
Try to see if you can refactor the CSS:
- maybe the same is repeated some places
- maybe the same element is styled in different ways and only one of the ways takes effect
- maybe there are styling properties that just do nothing, do not affect the design at all

Analyse the CSS as a detective and see if you can optimise at least a little bit 🕵🏽

portfolio_page.html
```diff
@@ -0,0 +1,117 @@
...
+
+    <div class="nav-bar">
+       <div class="logo">
+            <div class="circle">
+                <a href="#">
+              <span class="initial">R</span></a>
+            </div>
+            <p class="tagline">Designer & Developer</p>
+          </div>
+        <div class="nav-div" >
```
shpomp, line 21: Use semantic elements as much as possible, for example, for navigation.

portfolio_page.html
```diff
@@ -0,0 +1,117 @@
...
+              <img src="/material-symbols_developer-mode.png" width="60" height="60" viewBox="0 0 60 60" fill="none">
+              
+              <p class="card-text">Developer</p>
+            </div>
+            <p class="card-description">I have a passion for creating clean, maintainable code and developing functional and user-friendly applications.</p>
+          </div>
+        </div>
+      </div>
+
+      <div class="projects-container" id="projects">
```
shpomp, line 76: Here the classes are very well named! Very clear and structural 👍🏽

portfolio_page.html
```diff
@@ -0,0 +1,117 @@
...
+          Do you have something else on your mind?<br>
+          If you found my work interesting, please feel free to <span id="contact-text">contact</span> me!
+        </p>
+        <a href="mailto:rumanaraafi@gmail.com"class="contact-cta-button">Email me</a>
+      </div>
+    </div>
+  <!-- </div> -->
+
+    <footer class="footer">
+      <p class="footer-text">Find me at</p><p class="linkedin-text"><a href="https://www.linkedin.com/in/rumana-aafrin-abdul-huq-345135129/">LinkedIn</a></p>
```
shpomp, line 113: Try using a LinkedIn logo here that will be clickable. Will look nice and more attractive!

shpomp (changes requested):
Overall a very nice and simple page that has all the necessary parts. 👍🏽 
You already have some interesting eye-catching implementations which is very very good! 
A good job on the task itself, good well-structured markup, really nice work 👏🏽 ⭐ 

There are some comments on what to improve, some smaller things. 

Additionally, the files are in the wrong spot, so a step in the homework instructions is missed. 
Move the files to where they should be - the html-css folder. 🙏🏽

PR #48, opened 2023-08-24
git-week1/shahnawaz

shazkhan2:


2_git/week1/my_favorite_food.txt
```diff
@@ -0,0 +1 @@
+
```
paulafiaschi, line 1: Hi, this file is not necessary! Please, remove it and push again :)

paulafiaschi (changes requested):

2_git/week1/my_favorite_food.txt
shazkhan2: Hi Paula, I have pushed it again and the terminal shows that its being pushed but I dont see it on git. Can you?
Screenshot 2023-08-25 at 13 02 48

shazkhan2: Please check again

PR #52, opened 2023-08-24
git-week1/kateryna

katookap:
Added three txt files with favourite food and visited countries as described in exercise 1


Elisha2605 (approved):
Great job Kateryna! looks good!

PR #68, opened 2023-08-30
javascript1-week1/rumana

rumana-aafrin:
Evaluate my files

### DIFF
--- a/3_javascript/javascript1/week1/A dog age calculator.js
+++ b/3_javascript/javascript1/week1/A dog age calculator.js
@@ -2,8 +2,8 @@ const dogYearOfBirth= 2017;
 const dogYearFuture= 2027;
 const humanAge= dogYearFuture - dogYearOfBirth;
 const dogAge= 7*humanAge;
-let shouldShowResultInDogYears;
-if (shouldShowResultInDogYears= true){
+let shouldShowResultInDogYears=true;
+if (shouldShowResultInDogYears=== true){
  console.log( "Your dog will be " + dogAge + " dog years old in " + dogYearFuture );
 }
 else{
--- a/3_javascript/javascript1/week1/A dog age calculator.js
+++ b/3_javascript/javascript1/week1/A dog age calculator.js
@@ -1,0 +1,11 @@
+const dogYearOfBirth= 2017;
+const dogYearFuture= 2027;
+const humanAge= dogYearFuture - dogYearOfBirth;
+const dogAge= 7*humanAge;
+let shouldShowResultInDogYears;
+if (shouldShowResultInDogYears= true){
+ console.log( "Your dog will be " + dogAge + " dog years old in " + dogYearFuture );
+}
+else{
+    console.log( "Your dog will be " + humanAge + " human years old in " + dogYearFuture );
+}
--- a/3_javascript/javascript1/week1/A future age calculator.js
+++ b/3_javascript/javascript1/week1/A future age calculator.js
@@ -1,0 +1,4 @@
+const yearOfBirth= 1996;
+const yearFuture= 2050;
+const age= yearFuture - yearOfBirth;
+console.log("You will be " + age +"years old in " + yearFuture);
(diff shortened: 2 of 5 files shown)


3_javascript/javascript1/week1/A dog age calculator.js
```diff
@@ -0,0 +1,11 @@
...
+const dogYearFuture= 2027;
+const humanAge= dogYearFuture - dogYearOfBirth;
+const dogAge= 7*humanAge;
+let shouldShowResultInDogYears;
+if (shouldShowResultInDogYears= true){
+ console.log( "Your dog will be " + dogAge + " dog years old in " + dogYearFuture );
+}
+else{
+    console.log( "Your dog will be " + humanAge + " human years old in " + dogYearFuture );
+}
```
Pouryaak, line 11: 🐶 Your logic to calculate dog age and human age is spot on!
🚩 Be cautious with your if condition. You're using a single equals sign (=) which is an assignment, not a comparison. You'll want to use `===` to check for equality:
`if (shouldShowResultInDogYears === true)`
🔧 Currently, `shouldShowResultInDogYears` is not initialized to any value. It would be best if you initialize it to either true or false based on the desired behavior.

3_javascript/javascript1/week1/A future age calculator.js
```diff
@@ -0,0 +1,4 @@
+const yearOfBirth= 1996;
+const yearFuture= 2050;
+const age= yearFuture - yearOfBirth;
+console.log("You will be " + age +"years old in " + yearFuture);
```
Pouryaak, line 4: 👍 Great use of const for values that don't change!
💡 Consider using template literals to make the string concatenation more readable

3_javascript/javascript1/week1/A house price estimator.js
```diff
@@ -0,0 +1,25 @@
+// Peter
+let volumeInMetersPeter=8*10*10;
+let gardenSizeInM2Peter=100;
+let housePricePeter;
```
Pouryaak, line 4: 💡 To make this more structured, you might want to consider using an object to store each person's house details. As we can talked about it in the class, it's a good idea to group related values together in an object
🌱 Using objects can further help streamline your price-checking logic and make the code more modular and readable.

3_javascript/javascript1/week1/A house price estimator.js
```diff
@@ -0,0 +1,25 @@
...
+let gardenSizeInM2Julia=70;
+let housePriceJulia;
+housePriceJulia = volumeInMetersJulia * 2.5 * 1000 + gardenSizeInM2Julia * 300;
+console.log(housePriceJulia);
+if(housePriceJulia> 1.000000){
+ console.log("Julia's Pricing is high");
+}
+else{
+ console.log("Julia's Pricing is low"); 
+}
```
Pouryaak, line 25: 🚩 When comparing the prices, ensure you're using the correct values. Currently, you're comparing with `2.500000` which translates to 2.5. Use `2500000` instead.

3_javascript/javascript1/week1/Startup name generator.js
```diff
@@ -0,0 +1,26 @@
...
+"Core",
+"Nova",
+"Edge",
+"Ventures"
+];
+
+const startupName = firstWords.map((word, index) => word + " " + secondWords[index]);
+console.log(startupName);
+const totalCharacters = startupName.map(name => name.length);
+console.log("The startup: " + startupName[0] + " " + "contains" + " "+ totalCharacters[0] + " " + "characters");
```
Pouryaak, line 26: 💡 Consider using `const` for those variables you don't update, like `firstWords` and `secondWords`
👏 Nice approach in mapping firstWords and secondWords to create startup names.
🚩 However, for true randomness as per the initial instructions, think about incorporating `Math.random` as mentioned in the question.

Pouryaak (commented):
🚀 In Summary, you've made significant progress! Just a few refinements around structure and comparisons, and your code will be in top shape! Keep up the fantastic work!

3_javascript/javascript1/week1/Startup name generator.js
rumana-aafrin: Thankyou for pointing out the suggestions. I updated the changes now.

PR #73, opened 2023-08-31
javascript1-week1/andrii

ayaroshenkocom:
Here is my homework javascript1 week1

### DIFF
--- a/3_javascript/javascript1/week2/script.js
+++ b/3_javascript/javascript1/week2/script.js
@@ -25,6 +25,8 @@ console.log(fullname4);
 
 
 
+
+
 // Event application
 function getEventWeekday(daysUntilEvent) {
     const weekdays = ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"];
@@ -38,6 +40,8 @@ function getEventWeekday(daysUntilEvent) {
 
 console.log("The event will take place on " + getEventWeekday(3));
 console.log("The event will take place on " + getEventWeekday(7));
+
+
 
 
 
@@ -56,6 +60,7 @@ function recommendClothes(temperature) {
 
 const clothesToWear = recommendClothes(19);
 console.log(clothesToWear);
+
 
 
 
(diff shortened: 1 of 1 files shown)


3_javascript/javascript1/week1/script.js
```diff
@@ -0,0 +1,28 @@
+// Age-ify (A future age calculator)
+
+const yearOfBirth = 1987;
+const yearFuture = 2027;
+
+let age = yearFuture - yearOfBirth;
```
raihanmehran, line 6: using `const` instead of `let` would be better to go for, due to the fact that the value of `age` does not change.

3_javascript/javascript1/week1/script.js
```diff
@@ -0,0 +1,28 @@
...
+let dogYear = dogYearFuture - dogYearOfBirth;
+
+let shouldShowResultInDogYears = false;
+
+if (shouldShowResultInDogYears) {
+    dogYear *= 7;
+    console.log("Your dog will be " + dogYear + " dog years old in " + dogYearFuture);
+} else {
+    console.log("Your dog will be " + dogYear + " human years old in " + dogYearFuture);
+}
```
raihanmehran, line 25: Nice job 👍

raihanmehran (commented):
Overall, Nice Job 👍

3_javascript/javascript1/week1/script.js
```diff
@@ -25,4 +25,50 @@ if (shouldShowResultInDogYears) {
...
+    console.log("Peter is paying too much.");
+} else {
+    console.log("Peter is paying just right or less.");
+}
+
+if (realPriceJulia < costJulia) {
+    console.log("Julia is paying too much.");
+} else {
+    console.log("Julia is paying just right or less.");
+}
```
raihanmehran, line 60: Perfect 💯  solution!
One small suggestion, if you use `const` instead of `let`, as you can see value of the variables don't change. So, it's better to opt for `const`. This will help you to avoid accidental reassignment of these variables and also your code will be more readable.

3_javascript/javascript1/week1/script.js
```diff
@@ -25,4 +25,50 @@ if (shouldShowResultInDogYears) {
...
+
+let firstWords = ["Easy", "Awesome", "Corporate", "Brilliant", "Quantum", "Innovate", "Smart", "Dex", "Alpha", "Terra"];
+let secondWords = ["Solution", "Tech", "Corporation", "Dynamics", "Nexus", "Concepts", "Ideas", "Visions", "Cognition", "Nova"];
+
+let firstRandomIndex = Math.floor(Math.random() * 10);
+let secondRandomIndex = Math.floor(Math.random() * 10);
+
+let startupName = firstWords[firstRandomIndex] + " " + secondWords[secondRandomIndex];
+
+console.log("The startup: \"" + startupName + "\" contains" + startupName.length + "characters");
```
raihanmehran, line 74: One thing is clear, that code meets the requirements. That's a big plus 👍 

Meanwhile, there is still room for improvement.
- consider using `const` instead of `let`
- consider using template literals, which make your concise and more readable.

Example of using template literals:
`const firstName = 'John';`
`const lastName = 'Wick';`
console.log(\`My full name is ${firstName} ${lastName}`);

raihanmehran (approved):

3_javascript/javascript1/week1/script.js
ayaroshenkocom: Thank you very much :)

3_javascript/javascript1/week1/script.js
ayaroshenkocom: > using `const` instead of `let` would be better to go for, due to the fact that the value of `age` does not change.

Should I make changes or is this advice for the future?

3_javascript/javascript1/week1/script.js
ayaroshenkocom: Thank you for advice 👍

3_javascript/javascript1/week1/script.js
ayaroshenkocom: Oops, I forgot about template literals. Pourya talked about them in class. Indeed, it is much better to use it.

3_javascript/javascript1/week1/script.js
raihanmehran: Well, that’s up to you!
Think about it, if it makes sense make the changes.
But I would recommend making changes 😊

PR #81, opened 2023-09-06
javascript1-week1/nada

NadaJrad:


3_javascript/javascript1/week1/script.js
```diff
@@ -0,0 +1,72 @@
+let yearOfBirth = 1998;
+let yearFuture = 2023;
+
+// calculat age 
+let age = yearFuture - yearOfBirth; 
+let stringOutput = `You will be ${age} years old in ${yearFuture} .`
+// log age 
+console.log(stringOutput);
```
shpomp, line 8: Perfect! 
Very good naming of the result variable too 👍🏽

3_javascript/javascript1/week1/script.js
```diff
@@ -0,0 +1,72 @@
...
+let gardenSizeInM2 = 100;
+function housePrice(volumeInMeters,gardenSizeInM2){
+  return  volumeInMeters * 2.5 * 1000 + gardenSizeInM2 * 300;
+}
+//peter's house
+let pHeight = 10;
+let pWidth = 8;
+let PDepth = 10;
+let pVolumeInMeters = pHeight * pWidth * PDepth;
+let pGardenSize = 100;
```
shpomp, line 38: Not sure what "p" stands for in the variable names, maybe it is not necessary?

3_javascript/javascript1/week1/script.js
```diff
@@ -0,0 +1,72 @@
...
+  } else {
+    let dogAgeInHumanYears = dogYearFuture - dogYearOfBirth;
+    dogAgeStringOutput = `Your dog will be ${dogAgeInHumanYears} human years old in ${dogYearFuture}`;
+  }
+
+console.log(dogAgeStringOutput);
+// House price 
+
+let gardenSizeInM2 = 100;
+function housePrice(volumeInMeters,gardenSizeInM2){
```
shpomp, line 30: A more conventional what would be to call this getHousePrice (or caclulateHousePrice) - it is then instantly clear what this function is for. 
A name like housePrice is more suited for a variable that, unlike a function, does not _do_ anything.

3_javascript/javascript1/week1/script.js
```diff
@@ -0,0 +1,72 @@
...
+} else {
+  console.log(`Peter has payed ${peterHasPayed} the expected price for the house  ${peterHousePrice}.` );
+}
+
+// julis's house
+let jHheight = 8;
+let jWidth = 5;
+let jDepth = 11;
+let jVolumeInMeters = jHheight * jWidth * jDepth;
+let jGardenSize = 70;
```
shpomp, line 54: Okay, I can see now the "p" and "j" are for names, sorry :D
Ok for the homework, but not super informative in general so good to avoid in future. 
Either a different way of naming or the full name would be better for clarity 🙏🏽

3_javascript/javascript1/week1/script.js
```diff
@@ -0,0 +1,72 @@
...
+let juiliaHasPayed = 1000000;
+let  juiliaHousePrice = housePrice(jVolumeInMeters,jGardenSize)
+if (juiliaHasPayed < juiliaHousePrice ) {
+  console.log(`juilia has payed ${juiliaHasPayed} too little for the house ${juiliaHousePrice}.`);
+} else if (juiliaHasPayed > juiliaHousePrice) {
+  console.log(`juilia has payed ${juiliaHasPayed} too much for the house  ${juiliaHousePrice}.` );
+} else {
+  console.log(`juilia has payed ${juiliaHasPayed} the expected price for the house  ${juiliaHousePrice}.` );
+}
```
shpomp, line 63: As you can see, you are repeating a lot of the same logic in the calculations for Peter and Julia. 
How about extracting it into a function of its own?

3_javascript/javascript1/week1/script.js
```diff
@@ -0,0 +1,72 @@
...
+//Startup name generator
+
+const firstWords = ["Nice","Awosome","Corporate","Easy","sveart","Corporate","Example"];
+const secondWords = ["height","cold","Tech","car","soluations","name","startup"];
+const firstRandomNumber = Math.floor(Math.random() * 7);
+const secondRandomNumber = Math.floor(Math.random() * 6);
```
shpomp, line 69:
```suggestion
const secondRandomNumber = Math.floor(Math.random() * ?);
```
Challenge: if you wanted this to be dynamic, according to the array, what should you do? 💡

shpomp (changes requested):
Very good job, Nada! 
What is very important, you structure and name your code really well and in a clear way, which is very important and it is fantastic when it comes naturally as it does for you 👏🏽 ✨ 
Left some notes and comments and challenges, but overall a very good job ⭐

NadaJrad:
> Very good job, Nada! What is very important, you structure and name your code really well and in a clear way, which is very important and it is fantastic when it comes naturally as it does for you 👏🏽 ✨ Left some notes and comments and challenges, but overall a very good job ⭐

@shpomp, Thanks for your awesome feedbacks, it super motivating to hear, i will try always to do my best to become better,
i will go through your comments and do needed changes 💪

3_javascript/javascript1/week1/script.js
NadaJrad: i see, i think i will go with getHousePrice, it sounds clearer as you mentioned

3_javascript/javascript1/week1/script.js
NadaJrad: yes we can instead multiply by the Array.length to make it dynamic 🤩

PR #95, opened 2023-09-09
javascript1-week2/hiren

hirenkumar91:
# hi Need review for my Home work

### DIFF
--- a/3_javascript/javascript1/week2/script.js
+++ b/3_javascript/javascript1/week2/script.js
@@ -25,6 +25,11 @@ const fullname4 = getFullname("female", "Banjamin", "Hughes", true);
 const fullname1 = getFullname("male", "Banjamin", "Hughes", false);
 const fullname2 = getFullname("male", "Banjamin", "Hughes", true);
 
+console.log(fullname2);
+console.log(fullname1);
+console.log(fullname3);
+console.log(fullname4);
+
 /** Event application **/
 
 function getEventWeekday(daysFromToday) {
@@ -43,8 +48,9 @@ function getEventWeekday(daysFromToday) {
   return days[futureDate.getDate()];
 }
 
-console.log(getEventWeekday(9));
-console.log(getEventWeekday(2));
+const getEventday = getEventWeekday(9);
+
+console.log(getEventday);
 
 /** Weather wear */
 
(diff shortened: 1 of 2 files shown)


3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,129 @@
+/** Smart-ease - Goes Global! **/
+
```
szoboszlaianna, line 2: this should be `gender` with small g. Also consider using default values, so you don't have to defined them in each call. For example the parameters could look like this:
````
function getFullname(firstName, surName, gender="male",  useFormalName=False) {
...
}

```
Example use: 
````
const fullname3 = getFullname("Banjamin", "Hughes", "female");
const fullname4 = getFullname("Banjamin", "Hughes", "female", true);
const fullname1 = getFullname("Banjamin", "Hughes");
const fullname2 = getFullname("Banjamin", "Hughes", "male", true);

File: 3_javascript/javascript1/week2/script.js [hunk 2]
```diff
@@ -0,0 +1,129 @@
...
+function getFullname(Gender, firstName, surName, useFormalName) {
+  let fullName = "";
+
+  if (Gender === "female") {
+    if (useFormalName === true) {
```
szoboszlaianna, line 7: The if (useFormalName === true) condition can be simplified because useFormalName is already a boolean value. So you can simply write 
```
if(useFormalName) {
...
}
```

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,129 @@
+/** Smart-ease - Goes Global! **/
+
+function getFullname(Gender, firstName, surName, useFormalName) {
+  let fullName = "";
+
```
szoboszlaianna, line 5: You can simplify the code by checking for `useFormalName` first and then determining the title based on gender. This way, you can eliminate one condition block because the non-formal name is the same for both men and women.

```javascript
function getFullname(Gender, firstName, surName, useFormalName) {
  let fullName = "";

  // Check if formal name is requested
  if (useFormalName) {
    if (Gender === "female") {
      // Use "Lady" for a formal female title
      fullName = `Lady ${firstName} ${surName}`;
    } else {
      // Use "Lord" for a formal male title
      fullName = `Lord ${firstName} ${surName}`;
    }
  } else {
    // Use only the first and last name for informal names
    fullName = `${firstName} ${surName}`;
  }

  return fullName;
}
```

This code structure reduces redundancy and makes the logic more straightforward. If `useFormalName` is `true`, it will add the appropriate title based on gender, and if `useFormalName` is `false`, it will only use the first and last names for an informal name.

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,129 @@
...
+console.log(fullname2);
+console.log(fullname1);
+console.log(fullname3);
+console.log(fullname4);
+
+/** Event application **/
+
+function getEventWeekday(daysFromToday) {
+  const days = [
```
szoboszlaianna, line 36: I noticed a small issue in the `getEventWeekday` function where `getDate()` is used to determine the day of the week. However, `getDate()` returns the day of the month, which may lead to incorrect results.

To correctly determine the day of the week, the function should use `getDay()`. I recommend updating the code as follows:

```javascript
function getEventWeekday(daysFromToday) {
  const days = [
    "Sunday",
    "Monday",
    "Tuesday",
    "Wednesday",
    "Thursday",
    "Friday",
    "Saturday",
  ];
  const today = new Date();
  const futureDate = new Date(today);
  
  // Add the specified number of days to the current date
  futureDate.setDate(today.getDate() + daysFromToday);
  
  // Get the day of the week using getDay()
  return days[futureDate.getDay()];
}

const getEventday = getEventWeekday(9);

console.log(getEventday);

File: 3_javascript/javascript1/week2/script.js [hunk 5]
```diff
@@ -0,0 +1,129 @@
...
+  return days[futureDate.getDate()];
+}
+
+const getEventday = getEventWeekday(9);
+
+console.log(getEventday);
+
+/** Weather wear */
+
+function clothingChoice(temperature, humidity, windSpeed) {
```
szoboszlaianna, line 57: Nice work!

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,129 @@
...
+
+const class07Students = [];
+
+function addStudentToClass(studentName) {
+  if (studentName === "") {
+    console.log("Cannot add an empty string to the class.");
+  } else if (class07Students.includes(studentName)) {
+    console.log(`Student ${studentName} is already in the class.`);
+  } else if (
+    class07Students.length >= 6 &&
```
szoboszlaianna, line 102: Nice work!

szoboszlaianna (changes requested):

PR #96, opened 2023-09-09
javascript1-week2/Komal

270516:
I founded difficulty in student manager task. I tried but not getting the right result.

### DIFF
--- a/3_javascript/javascript1/week2/script.js
+++ b/3_javascript/javascript1/week2/script.js
@@ -1,7 +1,7 @@
 //First Task
 
 function getFullname(firstname, surname) {
-    return firstname + " " + surname;
+    return `${firstname} ${surname}`;
 }
 
 let fullName1 = getFullname("Komal", "Ranjan");
@@ -14,9 +14,9 @@ console.log(fullName2);
 
 function getFullname(firstname, surname, useFormalName) {
     if (useFormalName) {
-        return "Lady " + firstname + " " + surname;
+        return `${"Lady "} ${firstname} ${surname}`;
     } else {
-        return firstname + " " + surname;
+        return `${firstname} ${surname}`;
     }
 }
 let fullname1 = getFullname("Komal", "Ranjan", true);
@@ -29,9 +29,9 @@ console.log(fullname2);
 
 function getFullname(firstname, surname, useFormalName, title) {
     if (useFormalName && title) {
-        return title + " " + firstname + " " + surname;
+        return `${title} ${firstname} ${surname}` 
     } else {
-        return firstname + " " + surname;
+        return `${firstname} ${surname}`
     }
 }
 let maleFullname1 = getFullname("Pradeep", "Leuaa", true, "Lord");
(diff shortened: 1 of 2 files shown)


3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,144 @@
...
+
+let fullname2 = getFullname("Komal", "Ranjan", false);
+
+console.log(fullname1); 
+
+console.log(fullname2);
+
+function getFullname(firstname, surname, useFormalName, title) {
+    if (useFormalName && title) {
+        return title + " " + firstname + " " + surname;
```
szoboszlaianna, line 32: It is nicer to use string interpolation when we connect variables together. 
```javascript
`${title} ${firstname} ${surname}` 
```

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,144 @@
...
+}
+
+console.log(getEventWeekday(9)); 
+
+console.log(getEventWeekday(2)); 
+
+//Weather Wear
+
+function suggestClothings(temperature) {
+    if (temperature >= 30) {
```
szoboszlaianna, line 78: Nice work

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,144 @@
...
+
+function addStudentToClass(studentName) {
+    if (class07Students.length >= 6) {
+        console.log("Cannot add more students to class 07");
+    } else if (class07Students.includes(studentName)) {
+        console.log(`Student ${studentName} is already in the class`);
+    } else if (studentName === "Queen") {
+        class07Students.push(studentName);
+    } else if (studentName.trim() === "") {
+        console.log("Cannot add an empty string to the class");
```
szoboszlaianna, line 104: The problem is with this condition that if there are 7 students already in the list it will not get this far but will enter the first if condition: 
```
if (class07Students.length >= 6) {
    console.log("Cannot add more students to class 07");
  } 
```
To make sure that we check for the queen you can either move this up to be the first check. Or you can add it to the first check like this: 
```
//check if the list length is greater then 6 and it is not the queen.
if (class07Students.length >= 6 & studentName.toLowerCase() != "queen") {
    console.log("Cannot add more students to class 07");
  }
```
In order to catch these issues it is a good idea to add more than 7 students to the list and then the queen to test that it works.

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,144 @@
...
+
+//Second Task
+
+function getEventWeekday(daysFromToday) {
+    let today = new Date();
+
+    let weekdays = ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"];
+
+    let eventDayIndex = (today.getDay() + daysFromToday) % 7;
+
```
szoboszlaianna, line 61: this is unnecessary because the modulus returns the remainder which will be between 0 and 6. So it will never  be smaller then 0

szoboszlaianna (changes requested):

PR #97, opened 2023-09-09
javascript1-week2/kateryna

katookap:


3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,96 @@
...
+function addStudentToClass(studentName) {
+
+  if (class07Students.length >= 6) {
+    console.log("Cannot add more students to class 07");
+  } else if (studentName === "") {
+    
+    console.log("Student name cannot be empty");
+  } else if (class07Students.includes(studentName)){
+    console.log(`Student ${studentName} is already in the class`);
+ } else if (studentName.toLowerCase() === "queen") {
```
szoboszlaianna, line 80: The problem is with this condition that if there are 7 students already in the list it will not get this far but will enter the first if condition: 
```
if (class07Students.length >= 6) {
    console.log("Cannot add more students to class 07");
  } 
```
To make sure that we check for the queen you can either move this up to be the first check. Or you can add it to the first check like this: 
```
//check if the list length is greater then 6 and it is not the queen.
if (class07Students.length >= 6 & studentName.toLowerCase() != "queen") {
    console.log("Cannot add more students to class 07");
  }
```
In order to catch these issues it is a good idea to add more than 7 students to the list and then the queen to test that it works.

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,96 @@
...
+    
+    class07Students.push(studentName);
+  } else {
+    class07Students.push(studentName);
+  }}
+
+function getNumberOfStudents() {
```
szoboszlaianna, line 87: Nice work!

szoboszlaianna (changes requested):

PR #98, opened 2023-09-11
javascript1-week2/nada

NadaJrad:


3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,87 @@
+
+// exercice full_name
+function getFullname(firstname, surname, useFormalName, isWomen ){
+
+    if( firstname === "" & surname === ""){
```
szoboszlaianna, line 5: use `&&` You should use the logical AND operator (&&) for conditional checks instead of the bitwise AND operator.

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,87 @@
...
+        return "firstname and surname are missing"
+    }
+    if (useFormalName === true){
```
szoboszlaianna, line 8: Since `useFormalName` and `isWoman` are boolean you don't need the ===true comparison to check if they are true. You can simply call it like this 
```
if(useFormalName && isWoman)
{
...
}
```
Also the empty string `""`is a [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) value in javascript.So you don't need the comparison there either because they are converted to boolean and will be false. 
So we can have something like this:

```javascript
// Exercise: Full Name
function getFullname(firstname, surname, useFormalName, isWoman) {
    // Check if both firstname and surname are missing
    if (!firstname || !surname) {
        return "Firstname and surname are missing";
    }

    // Check if useFormalName is true
    if (useFormalName) {
        // Check if isWoman is true
        if (isWoman) {
            return `Lady ${firstname} ${surname}`;
        } else {
            return `Lord ${firstname} ${surname}`;
        }
    } else {
        // If useFormalName is false, return the informal full name
        return `${firstname} ${surname}`;
    }
}
```

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,87 @@
...
+
+const fullName = getFullname("Benjamin", "Hughes",true);
+console.log(fullName);
+let fullName1 = getFullname("nada", "jrad") ;
+let fullName2 = getFullname("jannah", "jrad") ;
+console.log(fullName1);
+console.log(fullName2);
+
+// event day 
+function getEventWeekday(numberOfTheDay){
```
szoboszlaianna, line 27: Nice!

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,87 @@
...
+   
+    return weekdays[(day + numberOfTheDay) % 7];
+}
+
+console.log(getEventWeekday(0));
+console.log(getEventWeekday(3));
+console.log(getEventWeekday(9));
+
+// weather
+function weather(tempertaure){
```
szoboszlaianna, line 40: nice!

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+let  clothesToWear = weather(8);
+console.log(clothesToWear);
+// Student manager
+const class07Students = [];
+function addStudentToClass(studentName) {
+    if (class07Students.length >= 6){
+       if(studentName === "Queen"){
+        class07Students.pop();
```
szoboszlaianna, line 56: Nice work. Although I think the Queen does not need to remove anyone to fit in. :) Also consider logging the return values because this way the user does not get feedback on what happened.

NadaJrad: @szoboszlaianna thanks a lot for the review and your feedback, i really appreciate it and it helps me to get better 🙏

PR #99, opened 2023-09-13
javascript1-week2/shahnawaz

shazkhan2:


3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,102 @@
+// Formal fullname
+function getFullname(firstname, surname, useFormalName, isWoman) {
+    if (firstname === "" || surname === "") {
```
szoboszlaianna, line 3: Since `useFormalName` and `isWoman` are boolean you don't need the ===true comparison to check if they are true. You can simply call it like this 
```
if(useFormalName && isWoman)
{
...
}
```
Also the empty string `""`is a [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) value in javascript.So you don't need the comparison there either because they are converted to boolean and will be false. 
So we can have something like this:

```javascript
// Formal fullname
function getFullname(firstname, surname, useFormalName, isWoman) {
    if (!firstname || !surname) {
        return `Please enter your full name`;
    }

    // Check if useFormalName and isWoman are both truthy
    if (useFormalName && isWoman) {
        return `Lady ${firstname} ${surname}`;
    }
    // Check if useFormalName is truthy and isWoman is falsy
    else if (useFormalName && !isWoman) {
        return `Lord ${firstname} ${surname}`;
    }
    else {
        return `${firstname} ${surname}`;
    }
}

File: 3_javascript/javascript1/week2/javascript.js [hunk 2]
```diff
@@ -0,0 +1,102 @@
+// Formal fullname
+function getFullname(firstname, surname, useFormalName, isWoman) {
```
szoboszlaianna, line 2: Nice work

3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,102 @@
...
+const fullName5 = getFullname("", "", false, false);
+
+console.log(fullName1);
+console.log(fullName2);
+console.log(fullName3);
+console.log(fullName4);
+console.log(fullName5);
+
+// Event application
+
```
szoboszlaianna, line 31: Nice and concise!

3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,102 @@
...
+    return weekdays[(today + numberOfTheDay) % 7];
+}
+
+console.log(getEventWeekday(0));
+console.log(getEventWeekday(1));
+
+// Weather wear
+// Create a function (that you have to name) that has temperature as parameter. 
+// Based on the temperature it should return a string with what the user should wear. 
+// You decide what the user should wear based on the temperature.
```
szoboszlaianna, line 45: Nice work

3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,102 @@
...
+
+
+// Student manager
+const class07Students = [];
+
+function addStudentToClass(studentName) {
+    if (class07Students.includes(studentName)) {
+        console.log(`Student ${studentName} is already in the class.`);
+    } else {
+        if (class07Students.length < 7 || class07Students.length > 6 && studentName === "Queen") {
```
szoboszlaianna, line 72: it is a good practice to use () to group together the conditions that belong together. (in this case the second 2).
This will still work, because in javascript the logical AND operator (&&) has higher precedence than the logical OR operator (||). This means that the expression inside the && is evaluated before the ||. But there could be cases where not grouping them will cause a different behavior.

szoboszlaianna (changes requested):

3_javascript/javascript1/week2/javascript.js
shazkhan2: I had a little issue with when I had "Queen" as a seperate condition. It was returning results even though the number of students was greater than 7. i then put it like this and then it started to behave :)

3_javascript/javascript1/week2/javascript.js
shazkhan2: Very useful information. I wouldnt have known it.

PR #105, opened 2023-09-16
javascript1-week3/meraj

MerajSharifi:


3_javascript/javascript1/week3/JS1-W3
```diff
@@ -0,0 +1,121 @@
...
+    return {
+      hours,
+      minutes,
+    };
+  }
+  
+  const { hours, minutes } = timeCalculator(travelInformation);
+  
+  console.log(`${hours} hours and ${minutes} minutes`);
+
```
Pouryaak, line 44: 🌟 I admire how you've approached the time calculation. This ensures you can get both hours and minutes without any overlap!
😊 Consider passing `travelInformation` as a parameter to the `timeCalculator()` function. This would make your function even more versatile and usable with different travel information.
📝 A tiny adjustment: For `travelTimeInHours`, it's more fitting to use `const` instead of `let` since the value isn't being reassigned.

3_javascript/javascript1/week3/JS1-W3
```diff
@@ -0,0 +1,121 @@
...
+      console.log(
+        `${series.title} took ${percentageOfLife.toFixed(2)}% of my life.`
+      );
+    }
+
+    console.log(`In total, that is ${totalMinutesWatched} minutes of my life.`);
+  }
+logOutSeriesText();
+
```
Pouryaak, line 87: 👍 Your strategy for computing the total time of each series in minutes is spot on. It's a great way to get an exact total.
⚠️ Just a heads up! Currently, the percentage calculation in your loop seems to cumulatively increase for each series, which might not give the correct percentage for individual series.
🧐 A quick tip for clarity: Some inline comments explaining certain values (like why 80 was used) can be super helpful for anyone reading the code.

3_javascript/javascript1/week3/JS1-W3
```diff
@@ -0,0 +1,121 @@
...
+  console.log(firstNote); 
+
+  /////log out 
+
+  function logOutNotesFormatted() {
+    for(i=0; i< notes.length-1; i++){
+    console.log(`The note with id: ${notes[i].id}, has the following note text: ${notes[i].content}`);
+    }
+  }
+   logOutNotesFormatted();
```
Pouryaak, line 121: 🌠 The `saveNote()` function is simple and effective! Brilliantly done!
🔍 There's a tiny hitch in the `getNote()` function. Instead of a single equals sign `(=)`, which assigns the value, you should use the comparison `(== or ===)`.
🤖 For your `logOutNotesFormatted()` function, there's a small oversight. Using `i < notes.length-1` might exclude the last note. Swapping it to `i < notes.length` should fix it.
🚀 Consistency Alert! Remember to use `let` or `const` when initializing the loop variable `i`.

Pouryaak (commented):
Your work on this PR is a testament to your coding journey. The manner in which you've structured and approached the problems exhibits clarity and diligence. While there are a few areas of improvement (which is absolutely natural and part of the growth process), the foundation is robust and promising. Keep this momentum going, embrace the feedback, and continue to shine in your coding endeavors. Excellent effort, and I'm eager to see where you'll take your projects next! 🌟🚀🎉

PR #113, opened 2023-09-17
javascript-javascript1-week3/NicoQ

lnknfan90:
Sorry for the late submission! Here is the HW for JS week 3. Thanks in advance for your review :)


shpomp:
image

PR #116, opened 2023-09-19
javascript1-week3/NicoQ

lnknfan90:
With the correct branch name 👍

### DIFF
--- a/3_javascript/javascript1/week3/Item array removal/index.html
+++ b/3_javascript/javascript1/week3/Item array removal/index.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <script src="script.js"></script>
+    <title>Item array removal</title>
+  </head>
+  <body></body>
+</html>
--- a/3_javascript/javascript1/week3/Series duration of my life/script.js
+++ b/3_javascript/javascript1/week3/Series duration of my life/script.js
@@ -1,0 +1,59 @@
+const seriesDurations = [
+  {
+    title: "Game of thrones",
+    days: 3,
+    hours: 1,
+    minutes: 0,
+  },
+  {
+    title: "Dragon Ball Z",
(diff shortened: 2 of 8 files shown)


3_javascript/javascript1/week3/Item array removal/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
shpomp, line 1: Neat file organization in the folders! 
Impressive systematic approach. 
There is a caveat, though, when trying to operate through your hw and f.x. run it with bash:
https://apple.stackexchange.com/questions/14683/how-to-cd-to-a-directory-with-a-name-containing-spaces-in-bash
Worth considering updating the naming convention 🙏🏽

3_javascript/javascript1/week3/Series duration of my life/script.js
```diff
@@ -0,0 +1,59 @@
...
+      series.days * 24 * 60 + series.hours * 60 + series.minutes;
+    totalMinutes = totalMinutes + seriesMinutes;
+  }
+  for (let i = 0; i < seriesDurations.length; i++) {
+    const series = seriesDurations[i];
+    const seriesMinutes =
+      series.days * 24 * 60 + series.hours * 60 + series.minutes;
+    const percentage =
+      (seriesMinutes / (lifeSpanInYears * minutesInYear)) * 100;
+    console.log(`${series.title} took ${percentage.toFixed(4)}% of my life`);
```
shpomp, line 51: Evidently, you are doing a lot of repetition here - same loop, some of the same actions. 
Optimize it to remove repetitive code, making sure you still get the same result 🙏🏽

shpomp (changes requested):
Overall very nice job! 👍🏽 
Good logic, good structure. Some optimization needed on one of the tasks 🤓

3_javascript/javascript1/week3/Item array removal/index.html
lnknfan90: Thanks! 
Regarding the naming, you are right, I had not thought about it, I will keep it in mind from now on :)

PR #124, opened 2023-09-23
javascript1-week4/NicoQ

lnknfan90:
Added 3 folders and 1 Txt in  week 4 Javascript1

Mi profile in CodeWars

https://www.codewars.com/users/LisekARG500/completed_solutions

### DIFF
--- a/3_javascript/javascript1/week4/CodeWars.txt
+++ b/3_javascript/javascript1/week4/CodeWars.txt
@@ -1,0 +1,3 @@
+Mi profile in CodeWars
+
+https://www.codewars.com/users/LisekARG500/completed_solutions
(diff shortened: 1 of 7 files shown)


shpomp (approved):
Without looking too much into detail, this looks ok at first glance. Possible chances for improvement for sure, but we will not spend time and energy on that now, for reasons that were discussed at length on Sunday sessions 😀
Rather than seeking new feedback on this PR, please go address the reviews on other weeks. You have very nice and detailed reviews there that look like they were disregarded. Perhaps one of them was addressed in a separate PR, not sure, won't try to investigate. 😬 
Either way, make sure to handle them ASAP and reach out if anything is unclear 🤝

PR #127, opened 2023-09-23
javascript1-week4/nada

NadaJrad:


3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,105 @@
...
+            user.todos.splice(index, 1);
+            return `Removed ${task} from your todo`;
+        } else {
+            return `${task} was not found in your todo list`;
+        }
+    }
+    // whats in my todo 
+    if (command.match("What is on my todo?")){
+        const numTodos = user.todos.length;
+        if (numTodos == 0) {
```
shpomp, line 44: It is ok to use == in this specific case, but it is worth remembering that there is a difference between == and ===. 
I would say it is safer to usually opt for ===, but it also depends on opinion a bit 😀 
Google around, but f.x. this is quite a good explanation:
https://www.scaler.com/topics/javascript/difference-between-double-equals-and-triple-equals-in-javascript/

3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,105 @@
...
+        const result = eval(question);
+        if (isNaN(result)){
+            return "Invalid question"
+        } else {
+            return result
+        }
+    }
+
+    const timerPrefix = "Set a timer for ";
+    const timerSuffix = " minutes";
```
shpomp, line 73: good naming for all the prefixes, indicative and logical! 👍🏽

shpomp (approved):
Just very nice and good job, @NadaJrad  There is nothing major to correct, just rest instead or use the time to work on previous homework / something else 😁 
👍🏽 ⭐ 💐

NadaJrad:
> Just very nice and good job, @NadaJrad There is nothing major to correct, just rest instead or use the time to work on previous homework / something else 😁 👍🏽 ⭐ 💐

thanks again karo, i am really glad to hear that i will keep trying to do my best.

3_javascript/javascript1/week4/script.js
NadaJrad: thanks i will read the article to understand  more he difference

PR #130, opened 2023-09-24
javascript1-week4/hiren

hirenkumar91:
hi my Home work for week 4 neew review

### DIFF
--- a/3_javascript/javascript1/week4/index.html
+++ b/3_javascript/javascript1/week4/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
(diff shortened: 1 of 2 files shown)


shpomp (approved):
@hirenkumar91 
Without looking too much into detail, this looks ok at first glance. Possible chances for improvement for sure, but we will not spend time and energy on that now, for reasons that were discussed at length on Sunday sessions 😀 
Rather than seeking new feedback on this PR, please go address the reviews on week1 and week2. You have very nice and detailed reviews there that were disregarded.
Make sure to handle them ASAP and reach out if anything is unclear 🤝

PR #140, opened 2023-09-28
javascript2-week1/eva

yagmureva:
Thanks for review.

### DIFF
--- a/3_javascript/javascript2/week1/script.js
+++ b/3_javascript/javascript2/week1/script.js
@@ -1,0 +1,82 @@
+//Find and count the Danish letters
+
+
+function countDanishLetters(inputString) {
+    
+    const letterCount = {
+      total: 0,
+      æ: 0,
+      ø: 0,
+      å: 0
+    };
+  
+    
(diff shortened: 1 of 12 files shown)


3_javascript/javascript2/week1/script.js
```diff
@@ -0,0 +1,82 @@
...
+      if (lowercaseString[i] === 'æ') {
+        letterCount.æ++;
+        letterCount.total++;
+      } else if (lowercaseString[i] === 'ø') {
+        letterCount.ø++;
+        letterCount.total++;
+      } else if (lowercaseString[i] === 'å') {
+        letterCount.å++;
+        letterCount.total++;
+      }
```
magdazelena, line 28: nice, but maybe you could think of the way how to access letters automatically so you could get rid of `if` statements?

magdazelena (changes requested):
why are all the other homeworks committed in the same PR?

You should be mindful to commit only needed changes per PR :) As a reviewer I don't know if that's a mistake or that you expect all these files to be reviewed

PR #143, opened 2023-09-28
javascript2-week1/kay

ksushaio:
completed JS2 week 1 homework


3_javascript/javascript2/week1/script.js
```diff
@@ -0,0 +1,128 @@
...
+  for (const char of lowercaseString) {
+    if (danishLetters.hasOwnProperty(char)) {
+      danishLetters[char]++;
+    }
+  }
+
+  const total = Object.values(danishLetters).reduce(
+    (acc, count) => acc + count,
+    0
+  );
```
magdazelena, line 21: nice use of reduce!

3_javascript/javascript2/week1/script.js
```diff
@@ -0,0 +1,128 @@
...
+document.addEventListener("DOMContentLoaded", function () {
+  const generateButton = document.getElementById("generateButton");
+  const spiritAnimalElement = document.getElementById("spiritAnimal");
+  const newAnimalButton = document.getElementById("newAnimalButton");
+  const eventSelector = document.getElementById("eventSelector");
+  const generateEventButton = document.getElementById("generateEventButton");
+
+  generateButton.addEventListener("click", function () {
+    generateSpiritAnimal();
+  });
```
magdazelena, line 47: if you have your function declared somewhere else already, you don't need to write callback like this. Just writing:
```
generateButton.addEventListener('click', generateSpiritAnimal)
```
would be enough. 

I would like to talk through it on the next class, please remind me if I forget to explain this 🙏

3_javascript/javascript2/week1/script.js
```diff
@@ -0,0 +1,128 @@
...
+
+  newAnimalButton.addEventListener("click", function () {
+    generateSpiritAnimal();
+  });
+
+  generateEventButton.addEventListener("click", function () {
+    const selectedEvent = eventSelector.value;
+    if (selectedEvent === "click") {
+      generateSpiritAnimal();
+    }
```
magdazelena, line 57: can you please explain why are you checking for the event value here?

magdazelena (changes requested):
I added some comments, but they are applicable in many places :) Feel free to ping me during the class so we can discuss some choices you made so we can talk through them

3_javascript/javascript2/week1/script.js
ksushaio: To be honest I asked chatGPT for some help (will try to avoid it from now on), I thought this was mandatory to check the event value. But now that you've raised the question, I see that there's no point in it. Also I realised I don't need  other `eventListeners` other than "click" ones. Thank you!

magdazelena: thank you for your honesty and I am glad you noticed <3 Can I screenshot your comment to prove my point to the class? 🙏

PR #146, opened 2023-09-30
javascript2-week1/Aaron

SpartanDeBoss:
HW for js2 week 1 :)

### DIFF
--- a/3_javascript/javascript2/week1/scripts.js
+++ b/3_javascript/javascript2/week1/scripts.js
@@ -1,0 +1,67 @@
+/// 1.2 Find and count the Danish letters ///
+
+function countDanishLetters(inputString) {
+  const danishLetters = {
+    æ: 0,
+    ø: 0,
+    å: 0,
+  };
+
+  for (const char of inputString) {
+    if (danishLetters[char]) {
+      danishLetters[char]++;
+    }
+  }
+  const total = Object.values(danishLetters).reduce(
+    (acc, count) => acc + count,
+    0
+  );
+  return { total, danishLetters };
+}
+
+/// Examples of what it returns ///
+const danishString = 'Jeg har en blå bil';
+countDanishLetters(danishString); // returns {total: 1, å: 1}
+
+const danishString2 = 'Blå grød med røde bær';
+countDanishLetters(danishString2); // returns {total: 4, æ: 1, ø: 2, å: 1}
+
+/// 2. Spirit animal name generator ////
+
+const nameInput = document.getElementById('nameInput');
+const generateButton = document.getElementById('generateButton');
+const spiritAnimalResult = document.getElementById('spiritAnimalResult');
+
+const spiritAnimalString = [
+  'The Strange Dog',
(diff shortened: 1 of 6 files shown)


3_javascript/javascript2/week1/scripts.js
```diff
@@ -0,0 +1,67 @@
...
+    0
+  );
+  return { total, danishLetters };
+}
+
+/// Examples of what it returns ///
+const danishString = 'Jeg har en blå bil';
+countDanishLetters(danishString); // returns {total: 1, å: 1}
+
+const danishString2 = 'Blå grød med røde bær';
```
szoboszlaianna, line 26: Make sure to console log the results to check if they are correct. I got `{ total: 0, danishLetters: { 'æ': 0, 'ø': 0, 'å': 0 }`for both of them

3_javascript/javascript2/week1/scripts.js
```diff
@@ -0,0 +1,67 @@
...
+function countDanishLetters(inputString) {
+  const danishLetters = {
+    æ: 0,
+    ø: 0,
+    å: 0,
+  };
+
+  for (const char of inputString) {
+    if (danishLetters[char]) {
+      danishLetters[char]++;
```
szoboszlaianna, line 12: it will never get here. The reason is that `danishLetters[char]` always returns the current value, which is 0 initially. And it is a `falsy` value in javascript (meaning it will always be false in the beginning.) Try to use the function `danishLetters.hasOwnProperty(char)`this will return true if the property exist.

3_javascript/javascript2/week1/scripts.js
```diff
@@ -0,0 +1,67 @@
...
+    const randomIndex = Math.floor(Math.random() * spiritAnimalString.length);
+    const spiritAnimalName = spiritAnimalString[randomIndex];
+    spiritAnimalResult.textContent = `Your new spirit animal is: ${userName} - ${spiritAnimalName}`;
+  } else {
+    spiritAnimalResult.textContent =
+      'Please enter your name before generating a spirit animal.';
+  }
+}
+nameInput.addEventListener('mouseenter', () => {
+  if (triggerEvent === 'inputHover') {
```
szoboszlaianna, line 63: trigger event is not defined. So the generate on hover does not work now.

szoboszlaianna (changes requested):

3_javascript/javascript2/week1/scripts.js
```diff
@@ -0,0 +1,67 @@
+/// 1.2 Find and count the Danish letters ///
+
+function countDanishLetters(inputString) {
```
szoboszlaianna, line 3: It would be nice to make it work with capital letters.

szoboszlaianna (changes requested):

SpartanDeBoss:
Hi Anna sorry for reaching out late but i have made all changes now.. thanks for reviewing it in the first place :)

PR #170, opened 2023-10-14
javascript2-week3/Rumana

rumana-aafrin:


3_javascript/javascript2/week3/The fastest presser in this realm.js
```diff
@@ -0,0 +1,41 @@
+let lCount = 0;
```
szoboszlaianna, line 1: Consider avoiding spaces in file names.  You could name it for example fastest_presser_game.js or something similar.

3_javascript/javascript2/week3/Function as a avariable.js
```diff
@@ -0,0 +1,22 @@
+    var array = [item1(), item2(), item3()];
```
szoboszlaianna, line 1: extrea features is an empty file?

3_javascript/javascript2/week3/exercise.js
```diff
@@ -0,0 +1,129 @@
...
+
+
+// 8. Double click
+document.addEventListener("dblclick", (e) => {
+  console.log( "double click!");
+});
+
+
+// 9.jokeCreator
+function jokeCreator (shouldTellFunnyJoke, logFunnyJoke, logBadJoke){
```
szoboszlaianna, line 115: Yes, you can simplify the `if` statement by directly using the condition without comparing it to `true`. 

```javascript
if (shouldTellFunnyJoke) {
    logFunnyJoke();
}
```
In JavaScript, the `if` statement checks the truthiness of the condition, so if `shouldTellFunnyJoke` is already a boolean, there's no need to compare it to `true`.

PR #173, opened 2023-10-14
javascript2-week1/Halima

halima770:
could you please review  my homework

### DIFF
--- a/3_javascript/javascript2/week1/script.js
+++ b/3_javascript/javascript2/week1/script.js
@@ -56,5 +56,51 @@ function generateSpiritAnimalName() {
 
 document.getElementById("generateButton").addEventListener("click", generateSpiritAnimalName);
 
+// 2.2. Setting up the events
+
+
+const generateButton = document.getElementById("generateButton");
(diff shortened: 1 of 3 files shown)


3_javascript/javascript2/week1/script.js
```diff
@@ -0,0 +1,65 @@
...
+];
+
+function generateSpiritAnimalName(name) {
+    const randomSpiritAnimal = spiritAnimals[Math.floor(Math.random() * spiritAnimals.length)];
+    const spiritAnimalName = name + " - " + randomSpiritAnimal;
+    document.getElementById('spiritAnimalName').textContent = "Spirit Animal Name: " + spiritAnimalName;
+}
+
+
+document.getElementById('newAnimalButton').addEventListener('click', function () {
```
shpomp, line 28: This formulation is not wrong, but it is not the most readable. 
Would be perhaps nice to declare the elements as variables at the top of the documents first and then add the event listeners as the next step 🙏🏽

shpomp (changes requested):
The homework is incomplete. It holds only on of several tasks.

PR #178, opened 2023-10-19
javascript3-week1/NicoQ

lnknfan90:
This HW was pretty cool 🤓 Thanks for your review!

### DIFF
--- a/3_javascript/javascript3/week1/Giphy/scripts.js
+++ b/3_javascript/javascript3/week1/Giphy/scripts.js
@@ -3,19 +3,17 @@ function domListener() {
   document.addEventListener("DOMContentLoaded", domLoader);
 }
 
+
 function domLoader() {
   const inputSearch = document.getElementById("search");
   const clickSearch = document.getElementById("click");
   const numberOfSearch = document.getElementById("quantity");
   const resultOfSearch = document.getElementById("result");
 
-  let toSearch = "";
-  let limitOfSearch = 25;
 
-  inputSearch.addEventListener(
-    "input",
-    () => (toSearch = inputSearch.value.toLowerCase())
-  );
+  let gifSearch = "";
+  let searchLimit = 25;
+  const APIKEY = "S6BbyLq63tmjQgRUkQ7T3wfVR0Jkr0Yy";
 
   clickSearch.addEventListener("click", parameterSearch);
   inputSearch.addEventListener("keydown", (event) => {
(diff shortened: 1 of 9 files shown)


3_javascript/javascript3/week1/Giphy/scripts.js
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function domLoader() {
+  const inputSearch = document.getElementById("search");
+  const clickSearch = document.getElementById("click");
+  const numberOfSearch = document.getElementById("quantity");
+  const resultOfSearch = document.getElementById("result");
+
+  let toSearch = "";
+  let limitOfSearch = 25;
```
shpomp, line 13: `toSearch` sounds neither like a variable nor a function.... since it is the search parameter. keyword, it could be named accordingly to make it clearer! 
`limitOfSearch` is okay, but it could be clearer by being more concise - searchLimit. 
What you named those is absolutely fine, I am just doing what's called "nitpicking" in code reviews 😀

3_javascript/javascript3/week1/Giphy/scripts.js
```diff
@@ -0,0 +1,53 @@
...
+
+  inputSearch.addEventListener(
```
shpomp, line 15: If I am not mistaken, this will run on every key entered, which there is no need to do. 
You need to take and lowercase and use the value only when a search is executed, so you could save resources by only doing this in the search execution function 💡

3_javascript/javascript3/week1/WeatherApp/scripts.js
```diff
@@ -0,0 +1,103 @@
...
+
+        const temperature = data.main.temp;
+        const weatherIconClass = selectWeatherIcon(temperature);
+        const weatherIcon = document.createElement("i");
+        weatherIcon.className = weatherIconClass;
+        weatherInfo.appendChild(weatherIcon);
+
+        windInfo(data);
+        cloudInfo(data);
+        dayTime(data);
```
shpomp, line 39: It might be more optimal to only pass the data that you need. 
Since you separated the functions so nicely, you could also make sure you don't throw all the data into them. 
But this is also maybe a nitpick :D

shpomp (changes requested):
Very good job! 
It is very nice and impressive to see how clean and well-structured your implementations are, there is very little to change or comment on. 
You are learning super well, bavo 👏🏽 ⭐ 

Huge apologies this stayed review-less for so long .😢

3_javascript/javascript3/week1/Giphy/scripts.js
lnknfan90: Thanks Karo! Super valuable review for me. I applied the changes :)

PR #185, opened 2023-10-21
javascript3-week1/Sevda

Sevda-NZ:
Sorry for delay! Thank in advance for review.


shpomp:
This has a lot of homework that should not be there. 
Make sure to only submit the relevant homework 🙏🏽

PR #187, opened 2023-10-21
javascript3-week1/Aaron

SpartanDeBoss:
JS3 Week 1 HW.. hope to get a review thanks in advanced


shpomp:
Submit only the needed homework. This includes previous homework. 
🙏🏽

SpartanDeBoss:
> Submit only the needed homework. This includes previous homework. 🙏🏽

My bad @shpomp it was a mistake, sorry but what will be the best way to do it? Delete the branch and resubmit or just delete the other files from it..  thanks for feedback!

shpomp:
> > Submit only the needed homework. This includes previous homework. 🙏🏽
> 
> My bad @shpomp it was a mistake, sorry but what will be the best way to do it? Delete the branch and resubmit or just delete the other files from it.. thanks for feedback!

- The best way would be to revert unneeded commits from this branch so that you keep this branch and only have the commits that are needed and learn to correct this mistake. In this case, try to search "how to revert a commit that has been pushed". That is the most proper way that you will need in the future.
- You could also try cherry-picking with git. In that case, try to search "how to cherry pick commits".

However, if it gets a bit much and confusing - that is normal. You could then try a manual simple workaround solution - resubmit the files in a new branch, for example. Deleting the files here is never a good idea as you would generally be trying to delete them from the whole repo, if you think about what a Pull Request is in essence. 

Try and see how it goes 😅 🙏🏽

SpartanDeBoss:
> > > Submit only the needed homework. This includes previous homework. 🙏🏽
> > 
> > 
> > My bad @shpomp it was a mistake, sorry but what will be the best way to do it? Delete the branch and resubmit or just delete the other files from it.. thanks for feedback!
> 
> * The best way would be to revert unneeded commits from this branch so that you keep this branch and only have the commits that are needed and learn to correct this mistake. In this case, try to search "how to revert a commit that has been pushed". That is the most proper way that you will need in the future.
> * You could also try cherry-picking with git. In that case, try to search "how to cherry pick commits".
> 
> However, if it gets a bit much and confusing - that is normal. You could then try a manual simple workaround solution - resubmit the files in a new branch, for example. Deleting the files here is never a good idea as you would generally be trying to delete them from the whole repo, if you think about what a Pull Request is in essence.
> 
> Try and see how it goes 😅 🙏🏽

hi @shpomp sorry for the late reply. i did try it , looks wired but I’m not sure IF it was the right thing tho 😂.. hopefully i can get a feedback thanks. 🙏🏾

PR #194, opened 2023-10-25
javascript1-week3/srushti

srushtisalke:


srushtisalke:
Accidentally push changes for JS1 week3. The homework was done before.

PR #196, opened 2023-10-26
javascript3-week2/NicoQ

lnknfan90:
Added the HW, thanks for your review  :)

### DIFF
--- a/3_javascript/javascript3/week2/CurrencyCalculator/script.js
+++ b/3_javascript/javascript3/week2/CurrencyCalculator/script.js
@@ -194,14 +194,14 @@ function domLoader() {
 
   function addListOfCurrencies() {
     currencies.forEach((currency) => {
-      var selectToCurrency = document.getElementById("toCurrency");
-      var newCurrencyTo = document.createElement("option");
+      const selectToCurrency = document.getElementById("toCurrency");
+      const newCurrencyTo = document.createElement("option");
       newCurrencyTo.value = currency;
       newCurrencyTo.text = currency;
       selectToCurrency.appendChild(newCurrencyTo);
 
-      var selectFromCurrency = document.getElementById("fromCurrency");
-      var newCurrencyFrom = document.createElement("option");
+      const selectFromCurrency = document.getElementById("fromCurrency");
+      const newCurrencyFrom = document.createElement("option");
       newCurrencyFrom.value = currency;
       newCurrencyFrom.text = currency;
       selectFromCurrency.appendChild(newCurrencyFrom);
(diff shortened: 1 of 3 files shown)


3_javascript/javascript3/week2/CurrencyCalculator/script.js
```diff
@@ -0,0 +1,240 @@
...
+    () => (toCurrency = toCurrencySelect.value)
+  );
+
+  const apiKey = "1bd1c44ffdf9e866b3f41d32";
+  let allData = "";
+  let fromCurrency = "EUR";
+  let toCurrency = "DKK";
+  let amount = 0;
+
+  const currencies = [
```
balance3840, line 31: You already have a list of all the currencies in the response of the API, so you could build this currency array from the API there. Visit this link for reference [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys)

3_javascript/javascript3/week2/CurrencyCalculator/script.js
```diff
@@ -0,0 +1,240 @@
...
+function domListener() {
+  document.addEventListener("DOMContentLoaded", domLoader);
+}
+
+function domLoader() {
+  const inputAmount = document.getElementById("amount");
+  const clickConvert = document.getElementById("convert");
+
+  clickConvert.addEventListener("click", getData);
+  inputAmount.addEventListener("input", () => (amount = inputAmount.value));
```
balance3840, line 11: Nice use of the callback functions here.

3_javascript/javascript3/week2/CurrencyCalculator/script.js
```diff
@@ -0,0 +1,240 @@
...
+          resolve(data);
+        })
+        .then(() => conversionResult());
+    });
+  }
+
+  function conversionResult() {
+    if (isNaN(amount)) {
+      const printResult = document.getElementById("conversionResult");
+      printResult.innerText = `Please enter the amount of money \n(use a period  "." as the decimal point)`;
```
balance3840, line 230: Nice idea to give feedback to the user when something goes wrong.

3_javascript/javascript3/week2/CurrencyCalculator/script.js
lnknfan90: Thanks for the suggestion!  I made the change taking the currencies from the API

PR #199, opened 2023-10-27
javascript3-week2/eva

yagmureva:
Thanks for review.


shpomp (changes requested):
Sorry it was not reviewed for long 😢 

But also, what is this? 😀 
image

All these changes definitely should not be there! 

The currency converter implementation is very very nice! 👏🏽 
I did not find much to comment on.

PR #203, opened 2023-10-28
javascript3-week2/komal

270516:
Thank you for review


3_javascript/javascript3/week2/script.js
```diff
@@ -0,0 +1,21 @@
+console.log("Let's start craziness")
+
+const convert = document.getElementById("convert");
```
balance3840, line 3: The API should populate all currency options instead of just EUR to DKK conversion. The code should be rewritten to comply with the homework.

PR #204, opened 2023-10-28
javascript3-week2/shahnawaz

shazkhan2:


3_javascript/javascript1/week4/script4.js
```diff
@@ -0,0 +1,104 @@
+let userName = [];
```
balance3840, line 1: Probably the name should be in plural ```userNames``` since it is an array.

3_javascript/javascript1/week4/script4.js
```diff
@@ -0,0 +1,104 @@
...
+
+        userName.push(name);
+
+        return `Nice to meet you ${name}`;
+    }
+
+    return "I'm sorry, I don't understand that command.";
+}
+
+if (command === "what is my name") {
```
balance3840, line 21: An idea here could be to convert to ```toLowerCase``` https://www.geeksforgeeks.org/javascript-string-tolowercase-method/.

This is just in case the user makes a mistake and writes ```What is my Name``` for example.

3_javascript/javascript1/week4/script4.js
```diff
@@ -0,0 +1,104 @@
...
+return `Removed ${removeItem} from your todo list`;
+}
+
+
+
+// what is on my todo
+
+function viewToDo() {
+    const excludedNames = ["Benjamin", "Ben"];
+    const activities = aboutMe.filter(activity => !excludedNames.includes(activity));
```
balance3840, line 46: Good idea to use ```filter``` here.

3_javascript/javascript3/week2/currencies.js
```diff
@@ -0,0 +1,163 @@
+const currencies = [
```
balance3840, line 1: You should get this currencies array from the API, since you already have that information in the response.

3_javascript/javascript3/week2/index.js
```diff
@@ -0,0 +1,72 @@
+document.addEventListener("DOMContentLoaded", function () {
+  const api = `https://v6.exchangerate-api.com/v6/${apiKey}/latest/USD`;
+  const fromDropDown = document.getElementById("from-currency");
+  const toDropDown = document.getElementById("to-currency");
+
+  function populateDropdowns() {
+    function addOption(currency, dropdown) {
```
balance3840, line 7: Nice idea to create ```addOption```function to avoid repetitive code.

3_javascript/javascript3/week2/index.js
```diff
@@ -0,0 +1,72 @@
...
+          console.error("Error:", error);
+          resultElement.innerHTML = "Error fetching exchange rates.";
+        });
+    } else {
+      alert("Please enter an amount to convert");
+    }
+  }
+
+  function setupEventListeners() {
+    document
```
balance3840, line 55: Good idea to add event listeners.

PR #212, opened 2023-11-04
javascript3-week2/nada

NadaJrad:


3_javascript/javascript3/week2/index.html
```diff
@@ -0,0 +1,30 @@
...
+  <div id="fromCurrencyContainer">
+    <p>From :</p>
+
+  </div>
+  <div id="toCurrencyContainer">
+    <p>To :</p>
+
+  </div>
+
+  <input type="button" value="Convert" id="mybtn">
```
babak-f, line 25: To make it easier to distinguish to elements, I would suggest you use a more descriptive ID, for example "convertBtn".

3_javascript/javascript3/week2/script.js
```diff
@@ -0,0 +1,68 @@
+
+const ApiKey = `https://open.er-api.com/v6/latest/EUR`;
```
babak-f, line 2: Since you are not using any variable inside the string, you can just use normal single quotes (').

3_javascript/javascript3/week2/script.js
```diff
@@ -0,0 +1,68 @@
...
+
+
+
+let toSelectList = document.createElement("select");
```
babak-f, line 6: Here you could use a const instead of let. Can you guess why?

3_javascript/javascript3/week2/script.js
```diff
@@ -0,0 +1,68 @@
...
+let fromSelectList = document.createElement("select");
+let rates;
+
+fetch(ApiKey)
+    .then(Response => Response.json())
```
babak-f, line 11: In JavaScript, variable names always start with a small letter. It won't affect the code working or not working, but it will help others understand your code easier :)

See https://www.w3schools.com/js/js_conventions.asp

3_javascript/javascript3/week2/script.js
```diff
@@ -0,0 +1,68 @@
...
+    .then(data => { //funA
+        rates = data.rates;
+        changeCurrency(data.rates)
+    });
+
+function changeCurrency(rates) {//funcB
+    //select list from 
+
+    const fromCurrency = document.getElementById("fromCurrencyContainer")
```
babak-f, line 20: Yess - here is the right use of cost :+1:

3_javascript/javascript3/week2/script.js
```diff
@@ -0,0 +1,68 @@
...
+    fromSelectList.id = "SelectFrom";
+    fromCurrency.appendChild(fromSelectList);
+
+    let key = Object.keys(rates)
+
+    for (var i = 0; i < key.length; i++) {
```
babak-f, line 26: In modern JavaScript `var` is not used anymore. See https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/

3_javascript/javascript3/week2/script.js
```diff
@@ -0,0 +1,68 @@
+
+const ApiKey = `https://open.er-api.com/v6/latest/EUR`;
+
+
+
+let toSelectList = document.createElement("select");
+let fromSelectList = document.createElement("select");
+let rates;
```
babak-f, line 8: Do you think you can move this declaration inside the function where `rates` is used?

In general, it is a good idea to try to avoid global variables as much as possible, since it becomes harder to control (in larger applications) what value they have at a given time.

babak-f (commented):
Great job :)

3_javascript/javascript3/week2/script.js
NadaJrad: yes because it is global variable 😀.

3_javascript/javascript3/week2/script.js
NadaJrad: yes thanks a lot for  helping me .

PR #216, opened 2023-11-08
databases-week1/yagmureva

yagmureva:
Thanks for review

### DIFF
--- a/5_databases/week1/homework
+++ b/5_databases/week1/homework
@@ -1,0 +1,55 @@
+-- Active: 1699216915341@@0.0.0.0@3306@hyf_lesson1
+
+SELECT COUNT(*)
+ FROM task;
+
+
+
+
+SELECT COUNT(*) 
+FROM task 
+WHERE due_date <= CURDATE();
+
+
+
+SELECT *
+FROM task
+WHERE status_id= 3;
+
+
+SELECT *
+FROM task
+WHERE status_id!= 3;
+
+
+
+SELECT *
+FROM task
+ORDER BY created DESC;
+
+
+SELECT MAX(created)
+from task;
+
+SELECT title, due_date
+FROM task;
+
+
+SELECT task.title, status.name
+FROM task, status
+
+
+
+
+SELECT status.name AS status_name, COUNT(task.id) AS task_count
+FROM status
+LEFT JOIN task ON status.id = task.status_id
+GROUP BY status.name;
+
+
+
+SELECT status.name AS status_name, COUNT(task.id) AS task_count
+FROM status
+LEFT JOIN task ON status.id = task.status_id
+GROUP BY status.name
+ORDER BY task_count DESC;


5_databases/week1/homework
```diff
@@ -0,0 +1,55 @@
...
+
+SELECT COUNT(*)
+ FROM task;
+
+
+
+
+SELECT COUNT(*) 
+FROM task 
+WHERE due_date <= CURDATE();
```
adionditsak, line 11: This is an awesome interpretation, that could actually be a real life solution. These tasks are from 2017 tho, so all of them are basically expired. Eventually just try just to look for what is NULL here to practice the `WHERE field IS NULL` statement

5_databases/week1/homework
```diff
@@ -0,0 +1,55 @@
...
+
+
+
+SELECT *
+FROM task
+WHERE status_id= 3;
```
adionditsak, line 17: Well done, but formatting is important in SQL, there should be a space before and after `=` so `WHERE status_id = 3;`

5_databases/week1/homework
```diff
@@ -0,0 +1,55 @@
...
+WHERE status_id!= 3;
+
+
+
+SELECT *
+FROM task
+ORDER BY created DESC;
+
+
+SELECT MAX(created)
```
adionditsak, line 31: This only returns the created field. Think more simple:

```
SELECT * 
FROM task 
ORDER BY created DESC LIMIT 1;
```

cool interpretation tho

5_databases/week1/homework
```diff
@@ -0,0 +1,55 @@
...
+from task;
+
+SELECT title, due_date
```
adionditsak, line 34: Good start, but we miss to search for the "database" string in title and description fields.

You do this with the `LIKE` operator:
https://www.w3schools.com/mysql/mysql_like.asp

5_databases/week1/homework
```diff
@@ -0,0 +1,55 @@
...
+FROM task;
+
+
+SELECT task.title, status.name
```
adionditsak, line 38: This returns the task 3 times with all statuses which means 103 rows of tasks, but we only have 35 tasks. We only want 1 task and its current status:

A little extra help for this:
FROM task
JOIN status
ON status.id=task.status_id;

5_databases/week1/homework
```diff
@@ -0,0 +1,55 @@
...
+
+
+SELECT status.name AS status_name, COUNT(task.id) AS task_count
+FROM status
+LEFT JOIN task ON status.id = task.status_id
+GROUP BY status.name;
+
+
+
+SELECT status.name AS status_name, COUNT(task.id) AS task_count
```
adionditsak, line 51: Well done!

adionditsak (changes requested):

adionditsak:
@yagmureva How is it going? Let me know if you need help :-)

PR #218, opened 2023-11-09
databases-week1/NicoQ

lnknfan90:
Thanks for your review  🤓

### DIFF
--- a/5_databases/week1/HW.sql
+++ b/5_databases/week1/HW.sql
@@ -1,0 +1,62 @@
+-- Active: 1699188605886@@0.0.0.0@3306@hyf_lesson1
+
+
+-- Task 1 Find out how many tasks are in the task table
+SELECT COUNT (title)
+FROM task
+
+-- Task 2 Find out how many tasks in the task table do not have a valid due date
+SELECT COUNT(*)
+FROM task
+WHERE due_date IS NULL
+
+-- Task 3 Find all the tasks that are marked as done
+SELECT *
+FROM task
+WHERE status_id = 3
+
+SELECT *
+FROM task JOIN status ON status.name = "Done" ;
+
+-- Task 4 Find all the tasks that are not marked as done
+SELECT *
+FROM task
+WHERE status_id != 3
+
+SELECT *
+FROM task JOIN status ON status.name != "Done" ;
+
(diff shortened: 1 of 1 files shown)


5_databases/week1/HW.sql
```diff
@@ -0,0 +1,62 @@
+-- Active: 1699188605886@@0.0.0.0@3306@hyf_lesson1
+
+
+-- Task 1 Find out how many tasks are in the task table
+SELECT COUNT (title)
```
adionditsak, line 5: Not a valid syntax, remove space between COUNT and `(title)`. Use `(*)` instead as we dont need to specify field

5_databases/week1/HW.sql
```diff
@@ -0,0 +1,62 @@
...
+SELECT *
+FROM task
+ORDER BY created
+LIMIT 1
+
+-- Task 7 Get the title and due date of all tasks where the title or description contains database
+
+SELECT title, due_date
+FROM task
+WHERE title LIKE "%database%"
```
adionditsak, line 44: Well done but you also have to check the description field for database. Do this by using the `OR` operator

5_databases/week1/HW.sql
```diff
@@ -0,0 +1,62 @@
...
+
+-- Task 8 Get the title and status (as text) of all tasks
+
+SELECT task.title, status.name
```
adionditsak, line 48: This returns the task 3 times with all statuses which means 103 rows of tasks, but we only have 35 tasks. We only want 1 task and its current status:

A bit of extra help on this one:
```
FROM task
JOIN status
ON status.id=task.status_id;
```

5_databases/week1/HW.sql
```diff
@@ -0,0 +1,62 @@
...
+
+-- Task 9 Get the name of each status, along with a count of how many tasks have that status
+
+SELECT status.name, COUNT(task.id) 
+FROM status JOIN task ON status.id = task.status_id
+GROUP BY status.name
+
+-- Task 10 Get the names of all statuses, sorted by the status with most tasks first
+
+SELECT status.name,  COUNT(task.id)
```
adionditsak, line 59: Well done

adionditsak (changes requested):

PR #219, opened 2023-11-10
databases-week1/komal

270516:
Thankyou for review

### DIFF
--- a/5_databases/week1/lesson1(HW).sql
+++ b/5_databases/week1/lesson1(HW).sql
@@ -1,0 +1,43 @@
+-- Active: 1699540080106@@0.0.0.0@3306@hyf_lesson1
+
+SELECT COUNT(*)
+FROM task;
+
+SELECT COUNT(*)
+FROM task
+WHERE due_date IS NULL;
+
+select task.title
+FROM task
+WHERE status_id = 3;
+
+SELECT task.title
+FROM task
+WHERE status_id != 3;
+
+SELECT *
+FROM task
+ORDER BY created DESC;
+
+SELECT *
+FROM task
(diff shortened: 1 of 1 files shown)


5_databases/week1/lesson1(HW).sql
```diff
@@ -0,0 +1,43 @@
...
+ORDER BY created DESC;
+
+SELECT *
+FROM task
+ORDER BY created DESC
+LIMIT 1;
+
+SELECT task.title, task.due_date
+FROM task
+WHERE title LIKE "%database%";
```
adionditsak, line 29: Well done but you also have to check the description field for database. Do this by using the `OR` operator

5_databases/week1/lesson1(HW).sql
```diff
@@ -0,0 +1,43 @@
...
+
+SELECT task.title, status.name
```
adionditsak, line 31: This returns the task 3 times with all statuses which means 103 rows of tasks, but we only have 35 tasks. We only want 1 task and its current status:

FROM task
JOIN status
ON status.id=task.status_id;

5_databases/week1/lesson1(HW).sql
```diff
@@ -0,0 +1,43 @@
...
+FROM task JOIN status;
+
+SELECT status.name , COUNT(task.id)
+FROM status JOIN task ON status.id = task.status_id
+GROUP BY status.name;
+
+SELECT status.name , COUNT(task.id)
```
adionditsak, line 38: Well done!

adionditsak (changes requested):

PR #225, opened 2023-11-11
Databases week1/aaron

SpartanDeBoss:


5_databases/week1/homework.sql
```diff
@@ -0,0 +1,60 @@
+-- Active: 1699628373418@@127.0.0.1@3306@hyf_lesson1
+
+-- 1.Find out how many tasks are in the task table
+SELECT count (*)
+FROM task;
+
+-- 2.Find out how many tasks in the task table do not have a valid due date
+SELECT count (*)
```
adionditsak, line 8: This is a nice interpretation of the task. Same problem as above with count

5_databases/week1/homework.sql
```diff
@@ -0,0 +1,60 @@
+-- Active: 1699628373418@@127.0.0.1@3306@hyf_lesson1
+
+-- 1.Find out how many tasks are in the task table
+SELECT count (*)
```
adionditsak, line 4: `count (*)` - space between count and what it is counting `(*)` is not a valid syntax and will return an error. Use `count(*)`

5_databases/week1/homework.sql
```diff
@@ -0,0 +1,60 @@
...
+
+
+-- 5.Get all the tasks, sorted with the most recently created first
+SELECT *
+FROM task
+ORDER BY created DESC;
+
+
+-- 6.Get the single most recently created task
+SELECT MAX (CREATED)
```
adionditsak, line 32: same problem with space as in count

adionditsak, line 32: This only returns the created field. Think more simple:


```
SELECT * 
FROM task 
ORDER BY created DESC LIMIT 1;
```

cool interpretation tho

5_databases/week1/homework.sql
```diff
@@ -0,0 +1,60 @@
...
+FROM task;
+
+
+-- 7.Get the title and due date of all tasks where the title or description contains database
+SELECT title, due_date
```
adionditsak, line 37: Good start, but we miss to search for the "database" string in title and description fields.

You do this with the `LIKE` operator:
https://www.w3schools.com/mysql/mysql_like.asp

5_databases/week1/homework.sql
```diff
@@ -0,0 +1,60 @@
...
+FROM task;
+
+
+-- 8.Get the title and status (as text) of all tasks
+SELECT task.title, status.name
```
adionditsak, line 42: This returns the task 3 times with all statuses which means 103 rows of tasks, but we only have 35 tasks. We only want 1 task and its current status:

A little extra help for this:
FROM task 
JOIN status 
ON status.id=task.status_id;

adionditsak (changes requested):

adionditsak:
@SpartanDeBoss How is it going, let me know if you need help?

PR #234, opened 2023-11-18
databases-week1/Sevda

Sevda-NZ:


Sevda-NZ:
It seemed good, so I closed it!

PR #242, opened 2023-11-23
databases-week3/Rumana

rumana-aafrin:


5_databases/week3/additional_queries.sql
```diff
@@ -0,0 +1,55 @@
...
+FROM meal
+WHERE price < 30;
+
+
+-- Get meals that still has available reservations
+SELECT meal.id, meal.title, meal.max_reservations, COUNT(reservation.id) AS reservations_taken
+FROM meal
+LEFT JOIN reservation ON meal.id = reservation.meal_id
+GROUP BY meal.id
+HAVING meal.max_reservations > COUNT(reservation.id) OR COUNT(reservation.id) IS NULL;
```
BJORN-VINTHER, line 12: Very nice approach, I didn't even know about `HAVING` 👍 

There is one issue with the query. Don't you think we should take into account how many people each reservation is made for? How would you do that?

5_databases/week3/additional_queries.sql
```diff
@@ -0,0 +1,55 @@
...
+
+
+-- Get meals that partially match a title. Rød grød med will match the meal with the title Rød grød med fløde
+SELECT title
+FROM meal
+WHERE title LIKE "Classic French%" ;
```
BJORN-VINTHER, line 18: How would you make it match with a meal named: `Good Ol' Classic French Fries`?

5_databases/week3/additional_queries.sql
```diff
@@ -0,0 +1,55 @@
...
+
+
+-- Get meals that has been created between two dates
+SELECT id,title,created_date
+FROM meal
+WHERE  meal.created_date >= ' 2023-01-01'
+AND meal.created_date <= '2023-11-05';
```
BJORN-VINTHER, line 25: Good solution which works. There is an even simpler syntax:
```sql
SELECT id, title, created_date
FROM meal
WHERE meal.created_date BETWEEN '2023-01-01' AND '2023-11-05';
```

5_databases/week3/additional_queries.sql
```diff
@@ -0,0 +1,55 @@
...
+
+
+-- Get only specific number of meals fx return only 2 meals
+SELECT * 
+FROM meal 
+LIMIT 0,2;
```
BJORN-VINTHER, line 31: You can simply do `LIMIT 2`

5_databases/week3/additional_queries.sql
```diff
@@ -0,0 +1,55 @@
...
+
+
+-- Get the meals that have good reviews
+SELECT meal.id, meal.title, AVG(review.stars) AS average_stars
+FROM meal
+JOIN review ON meal.id = review.meal_id
+GROUP BY meal.id, meal.title
+HAVING AVG(review.stars) >= 3
+LIMIT 2;
```
BJORN-VINTHER, line 40: Good solution. You can simply group on meal.id since it is a unique column. And no need to add the LIMIT 2 here.
```sql
SELECT meal.id, meal.title, AVG(review.stars) AS average_stars
FROM meal
JOIN review ON meal.id = review.meal_id
GROUP BY meal.id
HAVING AVG(review.stars) >= 3;
```

5_databases/week3/additional_queries.sql
```diff
@@ -0,0 +1,55 @@
...
+
+-- Get reservations for a specific meal sorted by created_date
+SELECT meal.id, meal.title, reservation.created_date, reservation.meal_id
+FROM meal
+JOIN reservation ON meal.id = reservation.meal_id
+WHERE meal.id=3
+ORDER BY reservation.created_date ASC;
```
BJORN-VINTHER, line 47: Your solution works. Since the task is only to get the reservations for a specific meal you can make the query simple as:
```sql
SELECT *
FROM reservation
WHERE meal_id = 3
ORDER BY created_date;
```

5_databases/week3/additional_queries.sql
```diff
@@ -0,0 +1,55 @@
...
+
+
+-- Sort all meals by average number of stars in the reviews
+SELECT meal.id, meal.title, AVG(review.stars) AS average_stars
+FROM meal
+JOIN review ON meal.id = review.meal_id
+GROUP BY meal.id, meal.title
+ORDER BY average_stars DESC;
```
BJORN-VINTHER, line 55: Good solution and great to see that you gave the output a fitting name using `AS average_stars`. 

You can just group by the meal.id as it is unique. 

A  user suddenly calls you and says that the result only shows 10 of her 12 meals. It seems the 2 missing have no reviews. How can we fix that so she will see all the meals?

BJORN-VINTHER (commented):
Great work Rumana. Very organized code. It is great that you divide the code into multiple files. It makes it easier for me to review. Also great solutions. I mostly only commented on minor syntax things.

PR #243, opened 2023-11-24
databases-week3/komal

komalranjan9:


5_databases/week3/datamodal.sql
```diff
@@ -0,0 +1,79 @@
...
+
+CREATE TABLE `review` (
+`id` int(10) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
+`title` VARCHAR(255) NOT NULL,
+`description` TEXT,
+`meal_id` int(10) UNSIGNED,
+`stars` int NOT NULL,
+`created_date` DATE,
+CONSTRAINT `fk_meal_review` FOREIGN KEY (`meal_id`) REFERENCES `meal` (`id`) ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```
BJORN-VINTHER, line 35: Looks great. It is good to see that you thought about which columns should be NULL or not. 👍 

It is a good practice to define for each column if the value should be NULL or NOT NULL. If not defined it will be NULL. However, the next developer in your code would have to know that in order to understand the code.

5_databases/week3/queries.sql
```diff
@@ -0,0 +1,133 @@
...
+
+    ---- Meal ----
+
+--Get all meals
+select *
+FROM meal;
+
+--Add a new meal
+INSERT INTO meal (id, title, description, location, `when`, max_reservations, price, reated_date)
+VALUES(3, "Naan", "Indian bread with garlic and butter", "Indian Royal", "2023-12-10 14:21:00", 5, "90", "2023-07-11");
```
BJORN-VINTHER, line 11: When inserting you most often don't want to explicitly insert the `id` as it is an `AUTO_INCREMENT` column. If you omit it, the database will automatically assign it the first available id. This way you don't have to figure out the next id to use and you also don't risk accidently assigning an id already in use.
```sql
INSERT INTO meal (title, description, location, `when`, max_reservations, price, reated_date)
VALUES("Naan", "Indian bread with garlic and butter", "Indian Royal", "2023-12-10 14:21:00", 5, "90", "2023-07-11");
```

5_databases/week3/queries.sql
```diff
@@ -0,0 +1,133 @@
...
+
+--Update a meal with any id, fx 1. Update any attribute fx the title or multiple attributes
+UPDATE meal
+SET title = "Shahi veg dum biryani", description = " Rice with royal indian spices and vegetables", price = "280"
+WHERE meal.id = 1;
+
+--Delete a meal with any id, fx 1
+DELETE meal
+FROM meal
+WHERE meal.id = 1;
```
BJORN-VINTHER, line 26: You typically just write 
```sql
DELETE FROM meal
WHERE meal.id = 1;
```

5_databases/week3/queries.sql
```diff
@@ -0,0 +1,133 @@
...
+WHERE price <90 ;
+
+--Get meals that still has available reservations
+SELECT *
+FROM meal
+WHERE max_reservations > (
+    SELECT COUNT(*)
+    FROM reservation
+    WHERE meal_id = meal.id
+);
```
BJORN-VINTHER, line 96: You have a great approach to this difficult query. 

Maybe we should take into account the number of guests for each reservation. How can we do that?

5_databases/week3/queries.sql
```diff
@@ -0,0 +1,133 @@
...
+FROM reservation
+WHERE meal_id = 3
+ORDER BY created_date;
+
+--Sort all meals by average number of stars in the reviews
+SELECT meal.*, AVG(review.stars) AS avg_stars
+FROM meal
+LEFT JOIN review ON meal.id = review.meal_id
+GROUP BY meal.id
+ORDER BY avg_stars DESC;
```
BJORN-VINTHER, line 129: Excellent solution and good that you chose to use a LEFT JOIN 💯

BJORN-VINTHER (commented):
Very good work Komal 🥇 You managed to solve a lot of complicated tasks while also keeping the code well organized and easy to read.

PR #246, opened 2023-11-25
Databases week3/andrii

ayaroshenkocom:
Please review my homework

### DIFF
--- a/5_databases/week3/queries.sql
+++ b/5_databases/week3/queries.sql
@@ -1,0 +1,104 @@
+----> Queries
+--> Meal
+SELECT *
+FROM meals
+
+
+INSERT INTO meals (title, description, location, meal_when, max_reservations, price, created_date)
+VALUES ('Shashlyk', 'Pork meat with onions on fire', 'Georgian restaurant', '2023-11-30 18:45:00', '14', '10.00', '2023-11-22')
+
+
+SELECT *
+FROM meals
+WHERE id = 1
+
+UPDATE meals 
+SET title = 'Kebab'
+WHERE id = 6;
+
+
+DELETE 
+FROM meals
+WHERE id = 1;
+
+--> Reservation
+
+SELECT *
+FROM reservations
+
+INSERT INTO reservations (number_of_guests, meal_id, created_date, contact_phonenumber, contact_name, contact_email)
+VALUES (3, 2, '2023-11-28', '555-555-5555', 'Alex', 'alex@example.com');
+
+SELECT * 
+FROM reservations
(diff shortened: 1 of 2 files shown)


5_databases/week3/queries.sql
```diff
@@ -0,0 +1,104 @@
...
+
+SELECT * 
+FROM meals
+WHERE price < 15
+
+SELECT meals.* 
+FROM meals
+JOIN reservations ON meals.id = reservations.meal_id
+WHERE reservations.created_date < meals.meal_when
+AND reservations.number_of_guests < meals.max_reservations;
```
BJORN-VINTHER, line 76: Good approach. It is great that you only considered reservations that were created before the meal takes place and also considered the number of guest per reservation to see if there are any available spots left.

However, you query would show duplicated meals if it has multiple reservations. You should probably try to add up the total number of guests instead of just the guests in each reservation. How can you modify the below query to do that?

```sql
SELECT ???
FROM meal
GROUP BY meal.id
```

5_databases/week3/queries.sql
```diff
@@ -0,0 +1,104 @@
...
+
+SELECT *
+FROM meals
+WHERE title LIKE 'Sushi%'
```
BJORN-VINTHER, line 80: A user suddenly mentions that her meal: `Delicious fried Sushi with Salmon` doesn't show up when she search using your query. What could be the issue?

5_databases/week3/queries.sql
```diff
@@ -0,0 +1,104 @@
...
+
+SELECT *
+FROM meals
+WHERE created_date BETWEEN '2023-11-20' AND '2023-11-25'
```
BJORN-VINTHER, line 84: Great usage of BETWEEN

5_databases/week3/queries.sql
```diff
@@ -0,0 +1,104 @@
...
+SELECT * 
+FROM reservations 
+WHERE meal_id = 2 
+ORDER BY created_date;
+
+SELECT meals.*, AVG(reviews.stars) AS avg_stars
+FROM meals
+LEFT JOIN reviews ON meals.id = reviews.meal_id
+GROUP BY meals.id
+ORDER BY avg_stars DESC;
```
BJORN-VINTHER, line 104: Perfectly done! Great that you use `AS` to give your output a fitting name. ⭐

BJORN-VINTHER (commented):
Great work Andrii. You managed to solve multiple complex queries perfectly <3

PR #247, opened 2023-11-25
databases-week3/eva

yagmureva:
Thanks for review

### DIFF
--- a/5_databases/week3/homework3
+++ b/5_databases/week3/homework3
@@ -1,0 +1,140 @@
+CREATE TABLE RESERVATION (
+  id INT UNSIGNED AUTO_INCREMENT NOT NULL PRIMARY KEY,
+  number_of_guests INT NOT NULL,
+  meal_id INT NOT NULL,
+  created_date DATETIME,
+  contact_phonenumber VARCHAR(255),
+  contact_name VARCHAR(255) NOT NULL,
+  contact_email VARCHAR(255) NOT NULL,
+  FOREIGN KEY (meal_id) REFERENCES Meal(id)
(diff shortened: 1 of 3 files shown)


5_databases/week3/homework3
```diff
@@ -0,0 +1,140 @@
...
+FROM meal
+-- Functionality
+-- 1.Get meals that has a price smaller than a specific price fx 90
+SELECT *
+FROM meal
+WHERE price < 90;
+-- 2.Get meals that still has available reservations
+SELECT *
+FROM meal
+WHERE max_reservations > 0;
```
BJORN-VINTHER, line 112: Here you only get the reservations with a max_reservations higher than 0. You need to take into account how many reservations have been created for this meal and how many people each reservation is for. How can you modify the below query to do that?
```sql

SELECT meal.id, meal.title, meal.max_reservations, SUM(reservation.???) AS reservations
FROM meal
GROUP BY meal.id
```

5_databases/week3/homework3
```diff
@@ -0,0 +1,140 @@
...
+SELECT *
+FROM reservation
+WHERE meal_id = 1
+ORDER BY created_date;
+-- 8.Sort all meals by average number of stars in the reviews
+SELECT meal.title, AVG(review.stars) AS average_stars
+FROM meal
+INNER JOIN review ON meal.id = review.meal_id
+GROUP BY meal.title
+ORDER BY average_stars DESC;
```
BJORN-VINTHER, line 140: The user tells you that not all meals appear in this output. It seems that only meals with a review are included. How can we fix it to include all meals?

BJORN-VINTHER (commented):
Good work Eva. Most of your SQL was spot on! Just make sure to pay extra attention to your queries modifying data if you wish to have a nice cozy Christmas 😉

PR #249, opened 2023-11-25
databases-week3/nada

NadaJrad:


5_databases/week3/homework3.sql
```diff
@@ -0,0 +1,36 @@
...
+
+CREATE TABLE `Review` (
+  `id` int(10) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `title` varchar(255) NOT NULL,
+  `description` text NULL DEFAULT NULL,
+  `meal_id` int(10) unsigned,
+  `stars` int,
+  `created_date` date,
+  CONSTRAINT `fk_Meal_Review` FOREIGN KEY (`meal_id`) REFERENCES `Meal` (`id`) ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```
BJORN-VINTHER, line 36: Good job. Pay attention to the `NULL` vs `NOT NULL`. If omitted, the column will be created as `NULL`. However, for some of your columns this is not ideal. I.e. what is a reservation without knowing which meal it belongs to?

5_databases/week3/Meal.sql
```diff
@@ -0,0 +1,31 @@
...
+VALUES(4,'Rød grød med','meduim','Denmark','2023-03-10','3','20','2023-09-10');
+--3 Get a meal with any id--
+SELECT*
+from `Meal`
+WHERE id = 1;
+
+--4 update a meal any id --
+UPDATE `Meal`
+SET id = 2
+WHERE id = 1;
```
BJORN-VINTHER, line 22: You generally never want to update a primary key of a row, i.e. `id`. The primary key should be set when the row is inserted in the database and then never changed. `id` is only used internally to make each row uniquely identifiable.

BJORN-VINTHER: I also think this was never a task. 😄

5_databases/week3/Meal.sql
```diff
@@ -0,0 +1,31 @@
...
+
+--5 update a meal any attribute --
+UPDATE `Meal`
+SET title = 'spaghetti'
+WHERE title = 'couscous';
```
BJORN-VINTHER, line 27: You query looks good however, in a real-life scenario always be careful when updating data like this. You have to be certain that you want to update ALL rows with this name, however many there are. If you have 1 row with title `couscous` then all is fine. But you could also have 200 rows with this name.
IF you need to update a specific entry it is generally better to use `WHERE id = ...` as we know there can only be 1 row with that id.

I think you may have misunderstood the task: _Update a meal with any id, fx 1. Update any attribute fx the title or multiple attributes_. 

```sql
UPDATE meal
set max_reservations = 8, price = 120
WHERE id = 1;
```

5_databases/week3/Reservation.sql
```diff
@@ -0,0 +1,29 @@
...
+--1--
+SELECT*
+FROM `Reservation`;
+--2--
+INSERT into Reservation (id, number_of_guests, meal_id, created_date, contact_phonenumber, contact_name, contact_email)
+VALUES(1, 10, 2, '2023-09-03', '+4528382639', 'nada_jrad', 'ada@gmail.com');
+INSERT into Reservation (id, number_of_guests, meal_id, created_date, contact_phonenumber, contact_name, contact_email)
+VALUES(2, 12, 3, '2023-12-02', '+4528452639', 'jannah', 'jannah@gmail.com');
+INSERT into Reservation (id, number_of_guests, meal_id, created_date, contact_phonenumber, contact_name, contact_email)
+VALUES(3, 12, 3, '2023-12-12', '+4528452639', 'jannah', 'jannah@gmail.com');
```
BJORN-VINTHER, line 11: You normally don't want to provide the `AUTO_INCREMENT` values like `id`. Instead omit that and let the database figure out the next free id and automatically assign it to your newly inserted row.

```sql
INSERT INTO Reservation (number_of_guests, meal_id, created_date, contact_phonenumber, contact_name, contact_email)
VALUES(10, 2, '2023-09-03', '+4528382639', 'nada_jrad', 'ada@gmail.com');
```

5_databases/week3/Additional-queries.sql
```diff
@@ -0,0 +1,44 @@
+use homework3;
+--1 Get meals that has a price smaller than a specific price fx 90--
+SELECT *
+FROM `Meal`
+WHERE price < 90;
+
+--2 Get meals that still has available reservations--
+SELECT Meal.*
+FROM Meal JOIN Reservation  ON Meal.id = Reservation.meal_id
+WHERE  Reservation.created_date > NOW();
```
BJORN-VINTHER, line 10: This is a hard query to make. You need to take into account the max_reservations and the total sum of guests of the related reservations. Try to finish the query below. 
```sql
SELECT meal.*, SUM(???) AS total_reservations
FROM meal
GROUP BY meal.id
```

5_databases/week3/Additional-queries.sql
```diff
@@ -0,0 +1,44 @@
...
+
+--3  Get meals that partially match a title. Rød grød med will match the meal with the title Rød grød med fløde-
+SELECT *
+FROM Meal
+WHERE title LIKE 'Rød grød med%';
```
BJORN-VINTHER, line 15: A user calls you and says that her meal `Vin og Rød grød med fløde` doesn't show when she searches for `Rød grød med`. What could be the issue?

5_databases/week3/Additional-queries.sql
```diff
@@ -0,0 +1,44 @@
...
+
+--4 meals between two date--
+SELECT *
+FROM Meal
+WHERE created_date BETWEEN '2023-01-01' AND '2023-12-31';
```
BJORN-VINTHER, line 20: Great usage of the BETWEEN

5_databases/week3/Additional-queries.sql
```diff
@@ -0,0 +1,44 @@
...
+
+--5 get limit of  number --
+SELECT *
+FROM Meal
+WHERE created_date BETWEEN '2023-01-01' AND '2023-12-31'
+LIMIT 1;
```
BJORN-VINTHER, line 26: Here we just ask you to get a specific number of rows. You can remove the WHERE clause.

5_databases/week3/Additional-queries.sql
```diff
@@ -0,0 +1,44 @@
...
+
+-- 6 good review --
+SELECT *
+FROM Review
+WHERE stars >= 3;
```
BJORN-VINTHER, line 31: Here you are asked to get the _Meals_ with a high scoring. You are returning the reviews. How can you return the meals instead?

5_databases/week3/Additional-queries.sql
```diff
@@ -0,0 +1,44 @@
...
+FROM Meal
+WHERE id = 2
+ORDER BY created_date;
+
+--8 AVG rating --
+SELECT Meal.*, AVG(Review.stars) AS average_stars
+FROM Meal
+LEFT JOIN Review ON Meal.id = Review.meal_id
+GROUP BY Meal.id
+ORDER BY average_stars DESC;
```
BJORN-VINTHER, line 44: Not an easy query to make but you solved it perfectly! Well done 🌟

BJORN-VINTHER (commented):
Good job Nada. You code is very well organized and easy to read. It is great that you divided the queries into a file for each category and gave each of them a comment explaining what they do. 
Overall, all your queries works as you wanted them to and pay attention to the details of the task. In the real world it can have a big impact on the solution you design.

PR #252, opened 2023-11-28
databases-week3/Sevda

Sevda-NZ:
Thanks for the review :)


5_databases/week3/mealSharing.sql
```diff
@@ -0,0 +1,200 @@
...
+SELECT  Meal.id, Meal.title, Review.title, Review.stars 
+FROM Meal, Review
+WHERE Review.stars > 4 AND Meal.id = Review.Meal_id;
+
+-- Get reservations for a specific meal sorted by created_date
+SELECT Meal.title, Reservation.Meal_id, SUM(Reservation.Number_of_guests), Reservation.Created_date AS Res_created_date
+FROM Reservation
+JOIN Meal ON (Meal.id=Reservation.Meal_id) WHERE 1
+GROUP BY Meal.title
+ORDER BY Reservation.created_date;
```
BJORN-VINTHER, line 192: It looks like there is some problems with your query. 
1. First of `WHERE 1` would always be true and is therefore redundant.
2. column names are all lowercase `some_sample_column`. 

In this case you can actually keep the query pretty simple as you are just asked to return the reservations for a specfic meal:
```sql
SELECT *
FROM Reservation
WHERE meal_id = 3
ORDER BY created_date;
```

5_databases/week3/mealSharing.sql
```diff
@@ -0,0 +1,200 @@
...
+
+-- Sort all meals by the average number of stars in the reviews
+SELECT Meal.*, AVG(Review.stars) AS avg_stars 
+FROM Meal
+JOIN Review ON Meal.id= Review.Meal_id
+GROUP BY Review.Meal_id
+ORDER BY avg_stars;
```
BJORN-VINTHER, line 199: Your user calls you and says that not all meals show up in the output. Especially newer meals seems to be missing 😮 What could be the issue and how do we fix it?

BJORN-VINTHER (commented):
Overall good job Sevda. You wrote some easy to understand queries and solved most of the tasks perfectly! Enjoy your Christmas knowing that you can now manage your own database  🎅

Sevda-NZ:
implemented the changes

PR #254, opened 2023-11-29
Warmup

ayaroshenkocom:


6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,22 @@
...
+    const numbers = args.map(Number);
+
+    const validNumbers = numbers.filter(num => !isNaN(num));
+
+    if (validNumbers.length > 0) {
+
+        const sum = validNumbers.reduce((acc, num) => acc + num, 0);
+        const average = sum / validNumbers.length;
+
+        console.log("Average:", average);
```
ddobby94, line 18: One tiny thing for improving string concatenation: 
```suggestion
        console.log(`Average: ${average}`);
```

ddobby94: But for the console.log() function, it doesn't really matter :)

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,22 @@
...
+
+const args = process.argv.slice(2);
+
+
+if (args.length === 0) {
+    console.log("No arguments provided. Please provide numbers as command line arguments.");
+} else {
+    const numbers = args.map(Number);
+
+    const validNumbers = numbers.filter(num => !isNaN(num));
```
ddobby94, line 11: Tiny thing: I usually like wrapping the incoming function property in brackets, so it is easier to add an extra param later if needed.
```suggestion
    const validNumbers = numbers.filter((num) => !isNaN(num));
```

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,22 @@
+// --> Warmup
+
+const args = process.argv.slice(2);
+
+
+if (args.length === 0) {
```
ddobby94, line 6: Very nice! We should always validate user inputs!!

6_nodejs/week1/avg.js
ddobby94: But this is a nice way of checking the data! Again good job for the validation!

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,22 @@
...
+    console.log("No arguments provided. Please provide numbers as command line arguments.");
+} else {
+    const numbers = args.map(Number);
+
+    const validNumbers = numbers.filter(num => !isNaN(num));
+
+    if (validNumbers.length > 0) {
+
+        const sum = validNumbers.reduce((acc, num) => acc + num, 0);
```
ddobby94, line 15: And this is how the `.reduce()` function should be used :)

ddobby94 (commented):
Good job! I only had very tiny comments! :)

PR #257, opened 2023-11-30
node-week1/nishadi

nishadipri:


6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,20 @@
+const args = process.argv.slice(2);
+
+if (!args.length){
```
babak-f, line 3: I would suggest to check if the length is at least 2, because otherwise taking the average would not make sense (if it's only one number).

babak-f (approved):
Great job :)

PR #258, opened 2023-12-02
node-week1/NicoQ

nicoquiroga90:
Thanks for the review

### DIFF
--- a/6_nodejs/week1/avg.js
+++ b/6_nodejs/week1/avg.js
@@ -1,14 +1,24 @@
-
-const args = require('minimist')(process.argv.slice(2));
+const args = require("minimist")(process.argv.slice(2));
 
 console.log(args.a);
 console.log(args.b);
 console.log(args.c);
 
-const Number1 = args.a
-const Number2 = args.b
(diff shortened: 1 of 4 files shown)


6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,14 @@
+
+const args = require('minimist')(process.argv.slice(2));
```
babak-f, line 2: What if the user enters 2 or less arguments?
In that case it will be good to show them a error message which guides them into the correct usage of the code.

babak-f: Hey
Sorry for the late reply
What I could think of is using the trim() function on the argument to remove the spaces, and then doing a boolean check (!true).
If the argument is only whitespace, trim() will turn it into an empty string and as a result it becomes false in the boolean check. 
Hope I understood your question correctly - or otherwise let me know :)

nicoquiroga90: Oh okey, I think I got it. Thanks for your help!

PR #262, opened 2023-12-08
node-week1/komal

komalranjan9:


6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,21 @@
...
+  
+  if (args.length === 0) {
+    console.log('No numbers provided. Please provide numbers as command line arguments.');
+  } else {
+    const numbers = args.map(Number);
+  
+    if (numbers.some(isNaN)) {
+      console.log('Invalid input. Please provide valid numbers as command line arguments.');
+    } else {
+      const average = calculateAverage(numbers);
```
ddobby94, line 17: Very nice way of using functions! :)

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,21 @@
+
+const calculateAverage = (numbers) => {
+    const sum = numbers.reduce((acc, num) => acc + Number(num), 0);
```
ddobby94, line 3: I don't think you need to convert `num` to a Number using the `Number()` function, since it is already a number
```suggestion
    const sum = numbers.reduce((acc, num) => acc + num, 0);
```

ddobby94 (commented):
Nice job! I only had a tiny comment :)

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,21 @@
...
+    return sum / numbers.length;
+  };
+  
+  const args = process.argv.slice(2);
+  
+  if (args.length === 0) {
+    console.log('No numbers provided. Please provide numbers as command line arguments.');
```
ddobby94, line 10: Great use of input validation!

PR #263, opened 2023-12-09
node-week2/nishadi

nishadipri:
Please review my homework

### DIFF
--- a/6_nodejs/week2/util.js
+++ b/6_nodejs/week2/util.js
@@ -1,0 +1,53 @@
+const documents = require("./document.json");
+
+exports.filterByQueryParam = (queryParam) =>{
+    return documents.filter(doc =>
+        doc.name?.includes(queryParam) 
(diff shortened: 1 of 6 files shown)


6_nodejs/week2/util.js
```diff
@@ -0,0 +1,53 @@
...
+    //         }
+    //     })
+    // })
+
+    return documents.filter(doc => {
+       let allFieldsMatch = true
+
+       for(const fieldName of fieldNames){
+        if(doc[fieldName]) {
+            switch (fieldName) {
```
babak-f, line 34: When it's only two cases, you could just use an if-else. Switch is usually for when there are more than two cases.

babak-f (approved):
Great job :+1:

ddobby94 (commented):
Nice work! Most of the files and functions were looking good, I had a few comments only about 1 specific function and some tiny comments :)

PR #264, opened 2023-12-09
Homework warmup NodeJS week2

ayaroshenkocom:


6_nodejs/week2/index.js
```diff
@@ -0,0 +1,94 @@
...
+app.get("/search", (req, res) => {
+    const query = req.query.q
+    const data = readFileSync('./documents.json');
+    const parsedData = JSON.parse(data);
+
+    if (!query) {
+        res.send(parsedData)
+    }
+
+    for (const document of parsedData) {
```
balance3840, line 23: Nice that you used a creative way to solve the problem. You could also consider using **.some** in the array, this way is much faster.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

6_nodejs/week2/index.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+});
+
+app.get("/documents/:id", (req, res) => {
+    const id = req.params.id
+    const data = readFileSync('./documents.json');
+    const parsedData = JSON.parse(data);
+
+    for (const document of parsedData) {
```
balance3840, line 42: You could also have used .**filter** here. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

6_nodejs/week2/index.js
```diff
@@ -0,0 +1,94 @@
...
+    const fields = req.body.fields
+    const data = readFileSync('./documents.json');
+    const parsedData = JSON.parse(data);
+
+    if (query && fields) {
+        res.status(400).json({ error: "Both query and fields can't be provided" });
+    }
+
+    if (query) {
+        for (const document of parsedData) {
```
balance3840, line 61: The same here with **.some**.

PR #267, opened 2023-12-16
node-week3/NicoQ

nicoquiroga90:
Thanks for the review

### DIFF
--- a/6_nodejs/week3/app.js
+++ b/6_nodejs/week3/app.js
@@ -1,0 +1,50 @@
+const knex = require("knex")({
+  client: "mysql2",
+  connection: {
+    host: process.env.DB_HOST || "0.0.0.0",
+    port: process.env.DB_PORT || 3306,
+    user: process.env.DB_USER || "root",
+    password: process.env.DB_PASSWORD || "pass",
+    database: process.env.DB_NAME || "hyf_node_week3_warmup",
+    multipleStatements: true, 
+  },
+});
+
+const express = require("express");
+const app = express();
+const port = process.env.PORT || 3000;
+
+app.use(express.json());
+
+const apiRouter = express.Router();
+app.use("/api", apiRouter);
+
+const contactsAPIRouter = express.Router();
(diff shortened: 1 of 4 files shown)


6_nodejs/week3/app.js
```diff
@@ -0,0 +1,50 @@
...
+        //query = query.orderByRaw(orderBy); the following request will drop the contacts table /api/contacts?sort=DROP TABLE contacts
+      query = query.orderByRaw(knex.raw('??', [orderBy]));
+    }
+  }
+
+
+  console.log("SQL", query.toSQL().sql);
+
+  try {
+    const data = await query;
```
babak-f, line 40: Shouldn't this be `query()`?

6_nodejs/week3/app.js
```diff
@@ -0,0 +1,50 @@
...
+    if (orderBy.length > 0) {
+        //query = query.orderByRaw(orderBy); the following request will drop the contacts table /api/contacts?sort=DROP TABLE contacts
+      query = query.orderByRaw(knex.raw('??', [orderBy]));
+    }
+  }
+
+
+  console.log("SQL", query.toSQL().sql);
+
+  try {
```
babak-f, line 39: I would suggest wrapping the entire function in a try-catch. That way, if an unexpected error happens while executing the rest of the function, the user will still get a meaningful error message.

babak-f (approved):

6_nodejs/week3/app.js
nicoquiroga90: Im not sure, because is calling a query _knex.select("*").from("contacts")_ and not a function. Anyway, that part of the code was wrote before, so im not 100% sure if I modify that I need to rewrite te rest of the code, if you guide me to do it would be great to don't break anything else 😊

6_nodejs/week3/app.js
nicoquiroga90: Perfect, I will do it ! :)

PR #281, opened 2024-01-09
react1-week1/meraj

MerajSharifi:


7_react/react1/week1/todo-list-app/src/TodoItem.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const TodoItem = (props) => {
+  return (
+    <div id="items">
+      <input type="checkbox" /> 
+      {props.description}, {props.deadline}
+    </div>
```
magdazelena, line 8: in general all good, but it's worth to keep in mind that regular rules of HTML and accessibility still apply in React, so I would push more towards using a list items :)

PR #283, opened 2024-01-09
react1-week1/rumana

rumana-aafrin:


7_react/react1/week1/src/App.js
```diff
@@ -0,0 +1,25 @@
+import './App.css';
+
+
+function MyLists(props){
+    const description= props.description;
+    const deadline= props.deadline;
+  return(
+    <ul>
+    <li>{description}, {deadline}</li>
+    </ul>
```
magdazelena: Please keep in mind that this way you are creating a new `ul` list for every list item in the App. Perhaps it would be worth to have only one `ul` in the App and here restrain yourself to `li`? Remember to update the component name then :)

rumana-aafrin: Yes! It makes sense now. I have updated the ul around the component name.

PR #284, opened 2024-01-09
react1-week1/nishadi

nishadipri:
Please review my homework. Thank you.


7_react/react1/week1/my-app/src/TodoItem/TodoItem.js
```diff
@@ -0,0 +1,19 @@
+import "./TodoItem.css"
+
+function TodoItem(props){
+
+    return(
+        <>
+        
+        
+            <ul className="todoItem">&#9733;{props.description},{props.deadline}</ul>
```
magdazelena: nice approach with using an object with data! However I think you mixed up `ul` and `li` tags, it should be the other way around :)

nishadipri: 🤣🤣 OMG yes I did.I will correct it.Thank you so much for the review.

PR #288, opened 2024-01-10
react1-week1/NicoQ

nicoquiroga90:
Thanks for your review 😊

### DIFF
--- a/7_react/react1/week1/react1-week1/src/App.js
+++ b/7_react/react1/week1/react1-week1/src/App.js
@@ -34,9 +34,9 @@ function Print(props) {
   const date = props.date;
 
   return (
-    <div>
+    <list>
       <span className="activity-color">{activity} </span> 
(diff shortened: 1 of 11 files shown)


7_react/react1/week1/react1-week1/src/App.js
```diff
@@ -0,0 +1,42 @@
...
+
+function Print(props) {
+  const activity = props.activity;
+  const date = props.date;
+
+  return (
+    <div>
+      <span className="activity-color">{activity} </span> 
+      <span className="date-color">{date}</span>
+    </div>
```
magdazelena: nice work, maybe it would be worth to consider that it is still a list and use HTML list instead of div for accessibility?

nicoquiroga90: Yes, you are right, I will modify it, thanks for your review :)

PR #291, opened 2024-01-11
react1-week1/komal

komalranjan9:


my-app/src/App.js
```diff
@@ -0,0 +1,42 @@
+
+import './App.css';
+
+const DateFormatter = ({ date }) => {
+  const formattedDate = new Date(date).toDateString();
+  return <span>{formattedDate}</span>;
+};
+
+const ToDoItem = ({description, deadline}) => {
```
BJORN-VINTHER, line 9: Wow buddy, we haven't even learned to use Props Destructuring yet. Nice one! ⭐

my-app/src/App.js
```diff
@@ -0,0 +1,42 @@
+
+import './App.css';
+
+const DateFormatter = ({ date }) => {
+  const formattedDate = new Date(date).toDateString();
+  return <span>{formattedDate}</span>;
+};
```
BJORN-VINTHER, line 7: Nice approach of formatting the days consistently. In the real world any dates will most likely be stored as `Date` rather than as a `string` in your array. There are many different ways to write down a date. I.e. `Wednesday Aug 22nd` `2023-08-22`, or `22/08/2023`. This can make it hard to handle in the code, thus we prefer to store it as a Date and use a function like yours to format it consistently in the UI. 

You can define it in the `todoItems` like a Date so you don't have to parse it from a string:
``` js
const todoItems = [
    {description: 'Get out of bed', deadline: new Date(2017, 9, 13) },
    {description: 'Brush teeth', deadline: new Date(2017, 9, 14) },
    {description: 'Eat breakfast', deadline: new Date(2017, 9, 15) },
  ];
```

BJORN-VINTHER (commented):
Very nice solution Komal. You are already ahead of our class. Great job! 🥇

PR #294, opened 2024-01-15
react1-week2/shahnawaz

shazkhan2:


7_react/react1/week2/my-app/src/App.css
BJORN-VINTHER, line 1: Wow, very nice css magic you do here. I didn't even know all of these cool tricks. Nice one!

7_react/react1/week2/my-app/src/App.js
```diff
@@ -0,0 +1,68 @@
...
+                Delete
+              </button>
+            </li>
+          );
+        })
+      ) : (
+        <div> No Items </div>
+      )}
+      </ul>
+    </div>
```
BJORN-VINTHER, line 66: Which components could you split the App into?

Make it a habit to think in components in React. Most of the power of React comes from dividing your code into logical & reusable components. It makes your code much easier to maintain as it grows, which it almost always will. If done properly it will also be much easier for other developers to understand the code.

7_react/react1/week2/my-app/src/App.js
```diff
@@ -0,0 +1,68 @@
...
+        list.map((listItem) => {
+          return (
+            <li key={listItem.id} className="list-item">
+              <input type="checkbox" id={listItem.id} />
+              <label for={listItem.id} className="strikethrough">
+                {listItem.description}
+              </label>
+              <button
+                onClick={() => {
+                  setList(list.filter((e) => e.id !== listItem.id));
```
BJORN-VINTHER, line 54: Nice and simple way of deleting an item. Good approach.

At a glance, it may not be obvious at a to all developers what `setList(list.filter((e) => e.id !== listItem.id))` does. You could consider to add a method to make the code easier to read for other developers and your future self:

``` js
const removeTodo = (todoId) {
    setList(list.filter((e) => e.id !== todoId))
}
```
then you can call it like this:
``` js
<button onClick={() => removeTodo(listItem.id)}/>
```

This is not a fixed rule but more of a personal preference you (and your team) would have to decide. 
* One developer could argue that the extra method would just bloat the code and be unnecessary 😞 
* Another developer always finds your code the easiest to understand and hail you as the best developer in the team. 🥇

BJORN-VINTHER (commented):
Good job Shahnawaz, you solved the task perfectly in terms of functionality. Try to split it into multiple components next time. 👍 

Also, I noticed that you accidently have both week1 and week2 homework in this PR. Try to make it a habit to always look remove any irrelevant files either before or after you raise your PR and remove those. Even in the industry it is a common mistake that you accidently committed some changes that should not be in PR. 

* At best, the PR reviewer will catch the mistake and ask you to remove those changes. 
* At worst, you accidently committed a local text file containing your bank information and login to your Netflix account. Not only is everyone in your company using your Netflix now, but they also created a Disney+ and HBO account using your credit card 😭

PR #295, opened 2024-01-18
react1-week2/NicoQ

nicoquiroga90:
Thanks for the review!


7_react/react1/week2/hw-react-week2/src/components/Container.jsx
```diff
@@ -0,0 +1,19 @@
+import { useState } from "react";
+
+import TaskList from "./TaskList";
+import FormTodo from "./FormTodo";
+
+const Container = () => {
```
BJORN-VINTHER: Try to give your component a more descriptive name if they are used for anything specific. `Container` doesn't say much about what it is used for. i.e. `TaskListContainer ` could be a more descriptive name.

7_react/react1/week2/hw-react-week2/src/components/Checkbox.jsx
```diff
@@ -0,0 +1,25 @@
+import React, { Fragment } from "react";
+import '../App.css';
+
+const Checkbox = props => {
+  const {
+    onChange,
+    data: { id, description, done }
+  } = props;
+  return (
+    <Fragment>
```
BJORN-VINTHER: `<Fragment>...</Fragment` is only needed if you need to return multiple elements.

There is also a shorthand version of it `<>...</>`
https://legacy.reactjs.org/docs/fragments.html

7_react/react1/week2/hw-react-week2/src/components/Container.jsx
```diff
@@ -0,0 +1,19 @@
...
+  const [list, setList] = useState([]);
```
BJORN-VINTHER: again, avoid overly generic names like `list`. It doesn't really say much about what the variable actually contains and can make the code harder to understand. What could be a more descriptive name?

7_react/react1/week2/hw-react-week2/src/components/FormTodo.jsx
```diff
@@ -0,0 +1,41 @@
...
+        <div className="taskInput">
+          <input
+            type="text"
+            className="textInput"
+            value={description}
+            onChange={(element) => setDescription(element.target.value)}
+          />
+          <button
+            className="addButtom"
+            disabled={description ? "" : "disabled"}
```
BJORN-VINTHER, line 31: Nice small detail of disabling the add button. Good job!

7_react/react1/week2/hw-react-week2/src/components/FormTodo.jsx
```diff
@@ -0,0 +1,41 @@
...
+    element.preventDefault();
+
+    handleAddItem({
+      done: false,
+      id: (+new Date()).toString(),
+      description,
+    });
+
+    setDescription("");
+  };
```
BJORN-VINTHER, line 17: Overall, I really like how you implemented this component. You generated a unique id by using the datetime, nice aproach. In the industry the backend would typically assign an id either as a number: `id: 123` or as a [unique id](https://www.uuidgenerator.net/dev-corner/javascript): `id: 'c32d8b45-92fe-44f6-8b61-42c2107dfe87'`

UUIDs are universal unique ids with an enormous number of unique combinations. it means that when you generate a random id it is practically impossible that your id has or will ever be generated before by any system ever made. 

In the words of Wikipedia:

> only after generating 1 billion UUIDs every second for the next 100 years, the probability of creating just one duplicate would be about 50%. The probability of one duplicate would be about 50% if every person on earth owns 600 million UUIDs

7_react/react1/week2/hw-react-week2/src/components/TaskList.jsx
```diff
@@ -0,0 +1,37 @@
...
+      {list.length ? (
+        <p>
+          <button className="DeleteButtom" onClick={onClickRemoveItem}>
+           {'Remove the selected task(s)'}
+          </button>
+        </p>
+      ) : null}
+    </div>
+  );
+};
```
BJORN-VINTHER: It is good that you challenged yourself and made your UI more advanced that what the homework required.  💪 

In the real world you have to be careful about freestyling too much, or at least make sure that you present your ideas with your manager or customer and make sure they agree on your ideas. Otherwise, it may bite you in the 🍑. 

Lets pretend I made a long Todo list for the entire month. I have more than 100 different todos and I am slowly making by way through the list and checking off a few todos per day. It really motivates me to see my progress. However, my nephew got hold of my tablet and added 5 todos of `play video games`. I want to remove these 5 todos but I don't want to lose my history of all the items I have finished so far. What should I do?

If this was not properly aligned with your manager, the error would fall on you unfortunately 😢

BJORN-VINTHER (commented):
Very nice solution. You have a good separation of components and the code is easy to understand. Good job Nico 👍 You have a good eye for details.

7_react/react1/week2/hw-react-week2/src/components/Container.jsx
nicoquiroga90: It makes sense, thanks for the suggestion. One question: it's straightforward now because I don't have many references to the name, and I can replace the names one by one. But is there any way for Visual Studio Code to modify all the references when I change the name of a component? thanks :)

7_react/react1/week2/hw-react-week2/src/components/Checkbox.jsx
nicoquiroga90: Yes, I was experimenting with fragments and then discovered there's a new format 🤓Thanks.

7_react/react1/week2/hw-react-week2/src/components/FormTodo.jsx
nicoquiroga90: Awesome, link -> fav

PR #303, opened 2024-01-20
react1-week2/rumana

rumana-aafrin:


7_react/react1/week2/src/App.js
```diff
@@ -0,0 +1,90 @@
...
+  {
+    id: 3,
+    description: "Eat Breakfast",
+    deadline: "Fri Sep 15 2017"
+  }
+];
+
+const MyLists = ({ items, check, handleCheck, deleteItem }) => (
+  <>
+    {items.map((props) => {
```
ddobby94, line 24: nit-pick:
`props` are usually referring to the React component props, and here we are iterating over an array using the `map(...)` function. So in the map function you can name your variables whatever you want, but you can name it something like:
`{items.map((arrayItem) => {`

7_react/react1/week2/src/App.js
```diff
@@ -0,0 +1,90 @@
...
+      const { id, description, deadline } = props;
+      return (
```
ddobby94, line 26: Very good, that you are using semantic html for the list, using `<ul>` and `<li>`!
A small thing here:
In this case it doesn't really makes sense to separate `<ul>` from the `<li>` so we could have everything here in React component:


```jsx
return (
  <ul>
    {items.map((props) => {
      <li key={id} className={`${check(id)} ${check(id) === "checked-item" ? "checked-item" : ""}`}>
        <input
          type="checkbox"
          onChange={() => handleCheck(id)}
          checked={check(id) === "checked-item"}
        />
        // ....
    })
  </ul>
```

7_react/react1/week2/src/App.js
```diff
@@ -0,0 +1,90 @@
...
+        <li key={id} className={`${check(id)} ${check(id) === "checked-item" ? "checked-item" : ""}`}>
+          <input
+            type="checkbox"
+            onChange={() => handleCheck(id)}
+            checked={check(id) === "checked-item"}
```
ddobby94, line 31: Nice way of using checking a function! 🎉 
We can modify the `isChecked` function a little bit, so the code becomes cleaner here :)
Check my other comment for `isChecked`

7_react/react1/week2/src/App.js
```diff
@@ -0,0 +1,90 @@
...
+  const isChecked = (id) => checked.includes(id) ? "checked-item" : "not-checked-item";
+  return (
+    <div className="App">
+      <h3>Todo List</h3>
+      {item.length === 0 ? (
+        <p>No items...</p>
+      ) : (
+        <ul>
+          <MyLists
+            check={isChecked}
```
ddobby94, line 79: The name is a bit confusing here. For function namings we can use verbs AND nouns, so it indicates what it actually does, something like: `isListItemChecked={isChecked}`

7_react/react1/week2/src/App.js
```diff
@@ -0,0 +1,90 @@
...
+    {items.map((props) => {
+      const { id, description, deadline } = props;
+      return (
+        <li key={id} className={`${check(id)} ${check(id) === "checked-item" ? "checked-item" : ""}`}>
+          <input
+            type="checkbox"
+            onChange={() => handleCheck(id)}
+            checked={check(id) === "checked-item"}
+          />
+          {description} {deadline} <button
```
ddobby94, line 33: We usually start a new line for each HTML component:
```jsx
{description} {deadline} 
<button
  type="button"
  onClick={() => deleteItem(id)}
>
  Delete
</button>
```

7_react/react1/week2/src/App.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+];
+
+const MyLists = ({ items, check, handleCheck, deleteItem }) => (
+  <>
+    {items.map((props) => {
+      const { id, description, deadline } = props;
+      return (
+        <li key={id} className={`${check(id)} ${check(id) === "checked-item" ? "checked-item" : ""}`}>
+          <input
```
ddobby94, line 28: For screen readers (people with limited vision use these to "see" websites), we should always have a `label` or an `aria-label` on input fields. 
More info on labels: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label

7_react/react1/week2/src/App.js
```diff
@@ -0,0 +1,90 @@
...
+    setItem((previousState) => [...previousState, newTask]);
+  };
+
+  const handleCheck = (id) => {
+    const updatedList = checked.includes(id) ? checked.filter((item) => item !== id) : [...checked, id];
+    setChecked(updatedList);
+  };
+
+  const deleteItem = (id) => {
+    const updatedItems = item.filter((item) => item.id !== id);
```
ddobby94, line 66: Nice way of using the `.filter` function!

7_react/react1/week2/src/App.js
```diff
@@ -0,0 +1,90 @@
...
+    setItem(updatedItems);
+  };
+
+  const isChecked = (id) => checked.includes(id) ? "checked-item" : "not-checked-item";
+  return (
+    <div className="App">
+      <h3>Todo List</h3>
+      {item.length === 0 ? (
+        <p>No items...</p>
```
ddobby94, line 75: Very good idea of having an empty state for the application! 

And here you can make the code cleaner with the empty-state management we can do this:
```jsx
  // ... Insert the same code from the component.
   
  // check for empty state
  if (item.length === 0) {
    // return the empty state here. This could be also a separate React.js component!
    return (
      <p>No items...</p>
    );
  }
  
  // display only the happy path
  return (
    <div className="App">
     ...

File: 7_react/react1/week2/src/App.js [hunk 9]
```diff
@@ -0,0 +1,90 @@
...
+              </button>
+        </li>
+        
+      );
+    })}
+  </>
+);
+
+function App() {
+  const [item, setItem] = useState(toDoLists);
```
ddobby94, line 47: nit-pick: the naming is a bit confusing here, for lists and arrays we usually use plural words like:
`items` or even better: `todoListItems`

7_react/react1/week2/src/App.js
```diff
@@ -0,0 +1,90 @@
...
+    const updatedList = checked.includes(id) ? checked.filter((item) => item !== id) : [...checked, id];
+    setChecked(updatedList);
+  };
+
+  const deleteItem = (id) => {
+    const updatedItems = item.filter((item) => item.id !== id);
+    setItem(updatedItems);
+  };
+
+  const isChecked = (id) => checked.includes(id) ? "checked-item" : "not-checked-item";
```
ddobby94, line 70: Actually we only need this part: 
`const isChecked = (id) => checked.includes(id)`

Since this function returns a boolean value, so that means the `checked={isChecked(id)}` will be translated to -> `checked={true}` or `checked={false}`

ddobby94 (commented):
Very nice job, you are using functions and React components correctly, I only had smaller comments to make the code more readable.
Please reach out to me if you have any questions!

PR #308, opened 2024-01-24
react1-week3/meraj

MerajSharifi:


7_react/react1/week3/todo-list-app/src/EmployeesList.js
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useState } from 'react';
+
+
+const Employees = () => {
+  const [data, setData] = useState([]);
+
+  const fetchData = async ()=>{
+
+    try {
+        let res = await fetch('https://gist.githubusercontent.com/benna100/5fd674171ea528d7cd1d504e9bb0ca6f/raw')
```
babak-f, line 10: Since `res` and `data` are not _modified_ later after being set, you can just use `const`. 
Of course it works this way too, but differentiating between `const` and `let` helps make the code easier to read.

7_react/react1/week3/todo-list-app/src/EmployeesList.js
```diff
@@ -0,0 +1,35 @@
...
+        let data = await res.json()
+        setData(data)
+    }catch(error){
```
babak-f, line 13: I would highly recommend getting into the habit of formatting your code before committing, so the code looks neater. 
You can even set VS code to to this automatically every time you save or after every few seconds. 

7_react/react1/week3/todo-list-app/src/EmployeesList.js
```diff
@@ -0,0 +1,35 @@
...
+        console.log("NOOOOOOOOOO")
```
babak-f, line 14: Is this error message for aliens? 😄

7_react/react1/week3/todo-list-app/src/EmployeesList.js
```diff
@@ -0,0 +1,35 @@
...
+    }
+}
+    
+    useEffect(() => {
+        fetchData()
+    }, [])
+
+    const onDelete = (employeesName) => {
+        setData((previousEmployees) => previousEmployees.filter((Employees) => Employees.name !== employeesName))
```
babak-f, line 23: Good job on the correct use of filter. 
Just remember that in JavaScript, variable names always start with a small letter. So it should be `employees` instead of `Employees`.
This applies to most other languages as well, and is primarily to distinguish between variables and classes.

7_react/react1/week3/todo-list-app/src/EmployeesList.js
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    return (
+        <div>
+        {
+            data.map(items => <p>{items.name}<button onClick={() => onDelete(items.name)}>delete me!</button></p>)
```
babak-f, line 29: Remember to assign the key to the child elements every time you use a map.

7_react/react1/week3/todo-list-app/src/TodoItem.js
```diff
@@ -0,0 +1,71 @@
+// TodoItem.js
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import Border from './Border'; 
+
+const TodoItem = ({ id, description, deadline, done, toggleDone, deleteTask, handleEditTask }) => {
```
babak-f, line 6: Neat and clear component structure. Good job 👍

7_react/react1/week3/todo-list-app/src/TodoList.js
```diff
@@ -0,0 +1,99 @@
+import React, { useState, useEffect } from 'react';
+import TodoHeader from './TodoHeader';
+import TodoItem from './TodoItem';
+
+const TodoList = () => {
```
babak-f, line 5: This component is also very well structured 👍

7_react/react1/week3/todo-list-app/src/TodoHeader.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const TodoHeader = () => {
```
babak-f, line 3: This is rather subjective and a matter of personal preference, but usually if a component is outputting just a text, it's not necessary to make it a separate component. I.e., you could just use the h2 tag whenever you need to use it. 

Of course, if it were to be more complex and have complex styling and so on, it could be defined as a separate component.

babak-f (approved):

PR #309, opened 2024-01-24
react1-week3/eva

yagmureva:
Thanks for review.

### DIFF
--- a/7_react/react1/week3/public/src/components/List.js
+++ b/7_react/react1/week3/public/src/components/List.js
@@ -1,0 +1,124 @@
+import React, { useEffect, useState } from "react";
+import ListItem from "./ListItem";
+import ReactDatePicker from 'react-datepicker';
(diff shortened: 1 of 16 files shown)


7_react/react1/week3/public/src/components/List.js
```diff
@@ -0,0 +1,124 @@
...
+    const newTask = {
+      id: Date.now(),
+      description: newTodo,
+      deadline: newDeadline,
+      done: false,
+    };
+
+    setList((previousState) => [...previousState, newTask]);
+    setNewTodo('');
+      setNewDeadline('');
```
babak-f, line 33: I would recommend getting into the habit of formatting your code before committing, so the code looks neater.
You can even set VS code to to this automatically every time you save or after every few seconds.

babak-f (approved):
Great job 👍

PR #313, opened 2024-01-30
react1-week3/shahnawaz

shazkhan2:


7_react/react1/week3/my-app/src/AddTodo.js
```diff
@@ -0,0 +1,41 @@
...
+      <br />
+      Deadline:
+      <input type="date" value={deadline} onChange={onChangeDeadline} />
+      <button
+        onClick={() =>
+          setTodos([
+            ...todos,
+            {
+              id: uuidv4(),
+              description: description,
```
babak-f, line 29: A minor thing: in JavaScript, when the two sides of an assignment have the same name, you can just write it once, i.e. `description, `.

babak-f (approved):
Great job
The components have a good and clear structure, so big kudos ;)

PR #314, opened 2024-01-30
react1-week3/NicoQ

nicoquiroga90:
Sorry for the delay! thanks for the review


7_react/react1/week3/hw-react-week3/src/components/Checkbox.jsx
```diff
@@ -0,0 +1,24 @@
+import "../App.css";
+
+const Checkbox = (props) => {
+  const {
```
babak-f: You could also directly "unpack" the props object in the line above to reduce the number of code lines, but of course this is also totally correct. 

See here: https://www.tutorialspoint.com/destructuring-of-props-in-reactjs

7_react/react1/week3/hw-react-week3/src/components/Checkbox.jsx
```diff
@@ -0,0 +1,24 @@
...
+    onChange,
+    data: { id, description, done },
+  } = props;
+  return (
+    <>
```
babak-f: Since this fragment (<>) has only one child, it can be omitted.

7_react/react1/week3/hw-react-week3/src/components/Container.jsx
```diff
@@ -0,0 +1,18 @@
+import { useState } from "react";
+import TaskList from "./TaskList";
+import FormTodo from "./FormTodo";
+
+const Container = () => {
+  const [list, setList] = useState([]);
+  const handleAddItem = (addItem) => {
+    setList([...list, addItem]);
```
babak-f, line 8: Good one

7_react/react1/week3/hw-react-week3/src/components/EditForm.jsx
```diff
@@ -0,0 +1,39 @@
...
+    onSave({
+      description: editedDescription,
+      dueDate: editedDueDate,
+    });
+  };
+
+  return (
+    <div className="editForm">
+      <form onSubmit={handleSubmit}>
+        <label>Description:</label>
```
babak-f: It's always a good idea to use the `for=` attribute on labels to connect them to the respective input.
This will help screen readers for people with visual impairment etc.
See here: https://www.w3schools.com/tags/tag_label.asp

7_react/react1/week3/hw-react-week3/src/components/TaskList.jsx
```diff
@@ -0,0 +1,133 @@
...
+      const updateList = list.map((item) => ({
+        ...item,
+        done: item.id === name ? checked : item.done,
+      }));
+      setList(updateList);
+    }
+  };
+
+  const onClickEditItem = (task) => {
+    setEditingTaskId(task.id === "" ? null : task.id);
```
babak-f: Have you accounted for `task.id` being undefined or null?
In general, for such checks, it is a good idea to use the Falsy concept in JavaScript (see https://developer.mozilla.org/en-US/docs/Glossary/Falsy).
Notice that this is very specific to JavaScript.

7_react/react1/week3/hw-react-week3/src/components/TaskList.jsx
```diff
@@ -0,0 +1,133 @@
...
+    setEditingTask(task);
+  };
+
+  const onSave = (updatedTask) => {
+    if (editingTaskId === null) {
```
babak-f, line 68: Same as above (null checking with Falsy)

7_react/react1/week3/hw-react-week3/src/components/TaskList.jsx
```diff
@@ -0,0 +1,133 @@
...
+    setList(updatedList);
+    setSelectedTasks([]);
+    setEditingTaskId(null);
+  };
+
+  const renderTaskList = list.map((item) => (
+    <div
+      key={item.id}
+      className={`taskRow ${item.done ? "done" : ""} ${
+        selectedTasks.some((task) => task.id === item.id) ? "selected" : ""
```
babak-f, line 96: I would suggest to move this check to its own constant outside the HTML markup, to make the class name easier to read and also to separate the logic part (JS) from the view part (HTML). 

It's a matter of style of course, and either way works fine.

7_react/react1/week3/hw-react-week3/src/components/TaskList.jsx
```diff
@@ -0,0 +1,133 @@
...
+    <div className="todoList">
+      {list.length ? renderTaskList : "Add your tasks"}
+      {editingTask && (
+        <EditForm
+          task={editingTask}
+          onCancel={() => setEditingTask(null)}
+          onSave={onSave}
+        />
+      )}
+      {list.length ? (
```
babak-f: You could replace this with `list.length && ...`, since the other side of the condition is null.

babak-f (approved):
Great job :)

PR #321, opened 2024-02-02
react2-week1/shahnawaz

shazkhan2:
The app can search a limited number of records due to API rate limits. The way to overcome it is to use a personal access token- ideally as an environment variable- and use the fetch function with this token.


7_react/react2/week1/my-app/src/Who.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+
+function WhoAreWe(){
+    return(
+    <div>
+    <h2 className='header'>
+    Who we are
+    </h2>
+<p>
```
babak-f, line 9: It is always a good idea to format the code before committing. It will also create a good impression in job interviews, code quizzes (for job applications) etc.

7_react/react2/week1/my-app/src/Context/useAppContext.js
```diff
@@ -0,0 +1,46 @@
+import React, { createContext, useContext, useState, useEffect, useCallback } from 'react';
+
+const AppContext = createContext();
+
+const AppProvider = ({ children }) => {
+  const [query, setQuery] = useState('');
+  const [data, setData] = useState([]);
+  const [loading, setLoading] = useState(false);
+
+  const fetchUsers = useCallback(async () => {
```
babak-f, line 10: Good job with the use of useCallback and also the error handling / try-catch block :+1:

babak-f (approved):

PR #322, opened 2024-02-03
react2-week1/NicoQ

nicoquiroga90:
thanks for review


7_react/react2/week1/user-search/src/App.js
```diff
@@ -0,0 +1,44 @@
...
+
+const App = () => {
+  const { users, loading, error, searchUsers, clearUsers } = useGithubContext();
+  const [query, setQuery] = useState('');
+
+  const handleSearch = (e) => {
+    const value = e.target.value;
+    setQuery(value);
+
+    if (value.trim() !== '') {
```
babak-f: You could remove the `!== ''` because an empty string is treated as false in JavaScript. See https://developer.mozilla.org/en-US/docs/Glossary/Falsy

7_react/react2/week1/user-search/src/components/GithubContext.jsx
```diff
@@ -0,0 +1,43 @@
...
+    setError('');
+
+    try {
+      const response = await fetch(`https://api.github.com/search/users?q=${query}`);
+      const data = await response.json();
+
+      if (response.ok) {
+        setUsers(data.items);
+      } else {
+        setError(data.message || 'Unknown error');
```
babak-f, line 21: Good style :+1:

babak-f (approved):

PR #328, opened 2024-02-09
react2-week2/shahnawaz

shazkhan2:


.env
```diff
@@ -0,0 +1,17 @@
+# Database - change to fit your database
+DB_HOST = 0.0.0.0
```
babak-f, line 2: Make sure to always add this file to git ignore. It should not be committed to the repository since it contains sensitive information (password etc)

6_nodejs/week3/app.js
```diff
@@ -0,0 +1,62 @@
...
+                return;
+            }
+        }
+    }
+  
+    const sql = query.toSQL().toNative();
+    console.log("SQL", sql);
+  
+    try {
+        const data = await query;
```
babak-f, line 52: I think the correct way to call query should be `query()`, i.e. with brackets. But I'm not fully sure about this - it might be one of the new JavaScript syntax features.

In general, for backward compatibility, I recommend using empty brackets whenever you make a function call that takes no parameters :)

7_react/react2/week1/my-app/src/Who.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+
+function WhoAreWe(){
+    return(
+    <div>
+    <h2 className='header'>
+    Who we are
+    </h2>
+<p>
+HackYourFuture is a not-for-profit coding school for people with limited access to education and the labor market. The majority of our students worldwide are refugees.
```
babak-f, line 10: Remember to always format your code before committing to the repository. You can set VS Code to do this automatically every time you save, for example.

7_react/react2/week3/Notetaker/src/pages/Notes.jsx
```diff
@@ -0,0 +1,44 @@
...
+import { v4 as uuidv4 } from "uuid";
+
+
+export default function Notes () {
+    const [notes, setNotes] = useState([])
+    const [search, setSearch] = useState("")
+    
+    const deleteNote = (id) => {
+        setNotes(notes.filter((n) => n.id !== id));
+
```
babak-f, line 14: Same note with code formatting :)

src/backend/api/meals.js
```diff
@@ -0,0 +1,268 @@
...
+
+//Week- 3
+
+//Returns all meals that are cheaper than maxPrice. Data type Number. api/meals?maxPrice=90
+
+router.get('/maxPrice', async (req, res) => {
+  try {
+    const maxPrice = req.query.maxPrice;
+
+    if (maxPrice !== undefined && isNaN(maxPrice)) {
```
babak-f, line 14: In JavaScript, it is always safer to check against both null and undefined. You can make use of the Falsy concept here and simply treat the value as a boolean: https://developer.mozilla.org/en-US/docs/Glossary/Falsy

The only thing to keep in mind is that 0 is also consider "false", so you may need to check for example like this: 
`if (maxPrice || maxPrice === 0)`

src/backend/api/meals.js
```diff
@@ -0,0 +1,268 @@
...
+
+// Week 2
+//GET - Returns all meals
+router.get("/", async (request, response) => {
+  try {
+    // knex syntax for selecting things. Look up the documentation for knex for further info
+    const allMeals = await knex("meal").select("*");
+    response.json(allMeals);
+  } catch (error) {
+    throw error;
```
babak-f, line 184: Return the error to the user with a descriptive error message and status :)
Notice that just throwing the error from catch makes the catching redundant - it was being thrown anyways.

src/backend/app.js
```diff
@@ -0,0 +1,43 @@
...
+
+const buildPath = path.join(__dirname, "../../dist");
+const port = process.env.PORT || 3000;
+const cors = require("cors");
+
+// For week4 no need to look into this!
+// Serve the built client html
+app.use(express.static(buildPath));
+
+// Parse URL-encoded bodies (as sent by HTML forms)
```
babak-f, line 20: Try to aim for clear spacing between the different blocks of the code. That way, it will look nicer in code interviews, quizzes etc.

babak-f (commented):
Great work

PR #353, opened 2024-07-08
databases-week2/Sevda

Sevda-NZ:
Please review HW/week 2 DB


5_databases/week2/part1.sql
```diff
@@ -0,0 +1,28 @@
+-- Add a task with these attributes: title, description, created, updated, due_date, status_id, user_id
+INSERT INTO task (id, title, description, created, updated, due_date, status_id) 
+VALUES (NULL, 'Become a developer', 'Work hard', '2017-10-06 04:03:52', '2017-09-20 19:34:43', '2017-12-22 20:58:03', 2);
```
NatsGt: You shouldn't add NULL to the id, you can omit this value by `INSERT INTO task (title...) VALUES ('Become a developer', ...)`. Since you stated that **ID** is AUTO_INCREMENT when you created the table, the value will be created automatically.

5_databases/week2/part1.sql
```diff
@@ -0,0 +1,28 @@
...
+
+-- Change a task status 
+UPDATE task
+SET status_id = 1
+WHERE id = 16; 
+
+-- Mark a task as complete
+UPDATE task
+SET status_id = 3
+WHERE id = 16;
```
NatsGt, line 24: Instead of hardcoding the status_id as 3, you should look for the id of status 'done' using a query.

5_databases/week2/part3.sql
```diff
@@ -0,0 +1,34 @@
+-- Active: 1699187035090@@127.0.0.1@3306@HYF_lesson2
+USE hyf_lesson2;
+
+--Get all the tasks assigned to users whose email ends in @spotify.com
+
+SELECT *
```
NatsGt: Here you should show the tasks of a user, not only the data of users.

NatsGt (changes requested):
Good job with your queries, but you need to correct some few. The diagram you created for your example is missing some tables. Additionally, you seem to be having some issue with your repository, you should fix this before merging.

Sevda-NZ:
The reviews are implemented 👍

